PR #36, opened 2022-01-08
ONDC hackathon - Product Pallet

smohan-dw:
A product pallet based on Stream to manage product lifecycle.


amarts (approved):
Looks like rating is not used right now. Other than that, it looks good :+1:

PR #54, opened 2022-02-21
add 'quantity' as an option in product listing

amarts:
Signed-off-by: Amar Tumballi <amar@dhiway.com>


amarts:
added product delegate as a method in the pallet, looks to be working fine so far.

PR #96, opened 2022-08-04
pallet meta & credit

smohan-dw:


Cargo.toml
```diff
@@ -28,8 +28,7 @@ members = [
 	"pallets/space",
 	"pallets/schema",
 	"pallets/stream",
-	"pallets/authorities",
-	"pallets/builder",
-	"pallets/foundation",
+	"pallets/anchor",
```
amarts: the word 'anchor' itself is bit confusing when i just check this. (as anchor is still a name we use for 'create' method in many pallets already). Need to document this better.

pallets/schema/src/lib.rs
```diff
@@ -151,77 +151,66 @@ pub mod pallet {
...
+				tx_schema.schema.controller.clone(),
+			)
+			.map_err(<Error<T>>::from)?;
+
+			SchemaDelegations::<T>::try_mutate(
+				tx_schema.identifier.clone(),
+				|ref mut delegation| {
+					ensure!(
+						delegation.len() + delegates.len() <= T::MaxSchemaDelegates::get() as usize,
+						Error::<T>::TooManyDelegates
```
amarts, line 196: Any hardcoded number here for 'Too Many"? Can we change this number in the future?

amarts: Applicable to every pallet where we use `TooManyDelegates` error.

amarts (commented):
WIP, but wanted to send comments as I get them.

Cargo.toml
amarts: `royalty_treasury`

amarts:
not sure why, the debug build had lesser dependency than the `--release` build, and also didn't start finalizing the blocks in `--dev` mode. Now building the release mode to run some tests, will update here after testing with release build.

amarts:
Ok, update on the above comment:

`cargo build --release` works. @smohan-dw confirmed with me that we support `--release` (with debug flags) and `--production` as two parameters for build. The default (`--debug`) is not supported in the build for now.

amarts (approved):

amarts (approved):

smohan-dw:
The --release build has a subsection for --dev, which adds debug symbols to the --dev execution.

PR #102, opened 2022-10-20
rebase and refactor

smohan-dw:


smohan-dw:
The `fast-runtime` feature configures certain runtime values via the feature flag.

- `fast-runtime` feature flag to toggle at build time whether the production or fast/testing values
- `prod_or_fast` macro to easily define runtime config values
- environment variables that can be set at build time to specify custom values

The feature can be enabled as follows:

`cargo build --release --features=fast-runtime` - defaults to 1sec Block time with 4 minute EPOCH

or with custom values:

`CORD_EPOCH_DURATION=40 CORD_LAUNCH_PERIOD=3 cargo build --release --features=fast-runtime`

PR #104, opened 2022-12-06
Pallet rating updates

smohan-dw:
Aggregation and value checks


amarts (commented):
Looks good. 

Only concern is, there is no more option for buyer app mapping here. (entity can be Gupta Store, and provider can be seller app is what I understand).

Is that ok? Also, are we making Apps do rating * 10 while sending the score?

amarts (approved):

PR #105, opened 2022-12-07
Pallet Score

smohan-dw:
Dependency - https://github.com/dhiway/cord.js/pull/30


pallets/score/src/lib.rs
```diff
@@ -0,0 +1,268 @@
...
+				&journal.entry.entity,
+			);
+
+			Self::deposit_event(Event::JournalEntry {
+				identifier,
+				entity: journal.entry.entity.clone(),
+				author: author.clone(),
+			});
+
+			Self::aggregate_score(&journal.entry);
```
amarts, line 241: Should we be doing the aggregate score before event?  normally the practise is Event is raised at the end when all tasks are complete... Or you have designed it to be like this.

runtime/src/lib.rs
```diff
@@ -363,7 +363,7 @@ impl pallet_balances::Config for Runtime {
 }
 
 parameter_types! {
-	pub const TransactionByteFee: Balance = 5 * MILLIUNITS;
+	pub const TransactionByteFee: Balance = 10 * NANOUNITS;
```
amarts, line 366: This should be documented. We have reduced the fees million times? :open_mouth:

amarts (approved):

amarts:
I see lot of conflicts in the repo. @smohan-dw one rebase was required i guess. 

Will push after resolving the changes! Please review.

PR #124, opened 2023-04-03
blocktime and epoch adjustments for testnet

smohan-dw:


amarts (approved):
intent is not clear in the PR. But the changes look fine.

smohan-dw:
The EPOCH configuration (5mins) was initially set to support dev environments. Changing that to 2 hours for Testnet.

PR #129, opened 2023-04-14
Test & benchmark updates

smohan-dw:
test benchmarking::benchmark_tests::test_benchmarks ... ok
test result: ok. 102 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 8.76s


smohan-dw:
- cargo test --all --all-targets --locked

with benchmark tests
 - cargo test --all --all-targets --locked --features runtime-benchmarks --no-fail-test

amarts (commented):
Looks good. Is there any numbers from the run?

smohan-dw:
Not yet. This PR only covers the benchmark updates and tests.

PR #139, opened 2023-04-23
Registry test & mock

rohibhan:
Registry pallet Test and Mock and lib changes for test cases.

- [x] Tested locally

### DIFF
--- a/pallets/registry/src/lib.rs
+++ b/pallets/registry/src/lib.rs
@@ -19,6 +19,12 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 #![allow(clippy::unused_unit)]
 
+#[cfg(any(test, feature = "runtime-benchmarks"))]
+mod mock;
+
+#[cfg(test)]
+mod tests;
+
 use frame_support::{ensure, storage::types::StorageMap, BoundedVec};
 
 pub mod types;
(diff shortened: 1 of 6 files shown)


pallets/registry/src/tests.rs
```diff
@@ -0,0 +1,294 @@
...
+				registry_id.clone(),
+				DID_00,
+			),
+			Error::<Test>::DelegateAlreadyAdded
+		);
+	});
+}
+
+#[test]
+
```
roynrishingha, line 260: I think there shouldn't be a new line between the attribute macro and function

pallets/registry/src/lib.rs
```diff
@@ -165,14 +171,15 @@ pub mod pallet {
...
 		/// Registry identifier is not unique
 		RegistryAlreadyAnchored,
 		/// Registry identifier not found
 		RegistryNotFound,
 		/// Only when the author is not the controller or delegate.
 		UnauthorizedOperation,
-		// Invalid Identifier
+		/// Invalid Identifier
 		InvalidIdentifier,
 		// Invalid Identifier Length
```
roynrishingha, line 184: Can we also make this also doc comment?

pallets/registry/src/tests.rs
rohibhan: Ya, this must be a typo will fix in next commit. Thanks

PR #169, opened 2023-05-22
Speed up CI build time by using 4-core machine

roynrishingha:
Current Github Action takes on average takes 1 hour. Trying to reduce the CI time by providing a better machine.
I'm hope it'll reduce build time and save us Action minutes.

### DIFF
--- a/.github/workflows/develop-pr.yml
+++ b/.github/workflows/develop-pr.yml
@@ -11,7 +11,7 @@ defaults:
 
 jobs:
   develop-pr-submit:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04-4-cores
     steps:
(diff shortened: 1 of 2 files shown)


amarts:
Here, its waiting for the worker for last 20mins. Can we try with 8 core or even 4 core? That may be good enough IMO.

smohan-dw:
How about using instances from AWS/GCP for this?

amarts:
We are not at a speed where we have more than 1 PR per hour, (70mins is current build time), so I feel optimizing here is not a thing we should be focusing on. This being an opensource project, no pressure on number of minutes too.

roynrishingha:
@amarts why 8 or 4 cores ? Why not 16 cores ?

Issue #177, opened 2023-05-30
About static cord binary's dependency


amarts:
Below is the error I got when I ran the cord binary build on ubuntu-22.04 (in ec2) on a debian machine on GCP.

```
amar@cord-test-1:~$ ./cord-static-x86-ubuntu key generate-node-key --file ./data/node-key
./cord-static-x86-ubuntu: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.30' not found (required by ./cord-static-x86-ubuntu)
./cord-static-x86-ubuntu: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by ./cord-static-x86-ubuntu)
./cord-static-x86-ubuntu: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ./cord-static-x86-ubuntu)
./cord-static-x86-ubuntu: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by ./cord-static-x86-ubuntu)
./cord-static-x86-ubuntu: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by ./cord-static-x86-ubuntu)
```

Highlight is, we surely need `libc` to be of same version on machines where it is build. Hence when we publish artefacts, we should make sure to name it properly with `arch` used, and also `libc` version.

roynrishingha:
Our current binary release from the Github action forces us to use ubuntu-22.04 OS on a x84_64 architecture. So to support any other architecture we can either use our docker image or release more binaries to cover other OS and Architectures.

NiranjanAP:
Closing this issue as it is duplicated with https://github.com/dhiway/cord/issues/226 with C4GT format

PR #178, opened 2023-05-30
Release CORD Arch x86 compiled binary from github action

roynrishingha:
- Upload the binary as a release asset
    - User should be able to `wget` or `curl` the pre-compiled binary

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,3 +48,56 @@ jobs:
 
           # Push the Docker image to the registry
           docker push ${{ env.DOCKER_SERVER }}/${{ env.DOCKER_USER }}/${{ env.IMAGE_REPOSITORY }}:${DOCKER_BRANCH}
+
+  x86_64-binary:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v3
+    
+      - name: Install dependencies
+        run: sudo apt-get update -y && sudo apt-get install -y protobuf-compiler libssl-dev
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -48,3 +48,56 @@ jobs:
...
+
+      - name: Upload Binary
+        id: upload_binary
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_path: target/release/cord
+          asset_name: cord-ubuntu-x86
```
amarts: can we add ubuntu version name ? (instead of latest, what we get here? like `22-04` or something like that).

Check #177 on the issues.

amarts (dismissed):

.github/workflows/release.yml
roynrishingha: Do you mean to name it like `cord-x86-ubuntu-22-04` ?

amarts: Yes, but add branch also at the end. So `cord-$branch-$arch-$os`

roynrishingha:
Should we run this on our `release-0.8.0` branch  and/or `develop` branch ?

amarts:
We release only on release branch. So, it should be '*' (if that is allowed). ie, whenever there is a release, pick the corresponding branch, and run the script.

amarts (approved):

PR #181, opened 2023-06-23
chore: release-v0.9.43

smohan-dw:
This update replaces/removes the following CLI parameters:
- `--rpc-max-payload` (replaced by `--rpc-max-request-size` and `--rpc-max-response-size`)
- `--ws-max-out-buffer-capacity`
- `--ws-external` (replaced by `--rpc-external`)
- `--unsafe-ws-external` (replaced by `--unsafe-rpc-external`)
- `--ipc--path`
- `--ws-port` (replaced by `--rpc-port`)
- `--ws-max--connections` (replaced by `--rpc-max-connections`)
- `--rpc-http` (replaced by `--rpc-addr`)
- `--rpc-ws` (replaced by `--rpc-addr`)


amarts:
@roynrishingha @Rithwik0480 Please note these changes. Specifically we use some of these options in deploy scripts. Good to keep an eye on when we make a release on this.

test-utils/src/lib.rs
```diff
@@ -179,14 +179,14 @@ pub async fn wait_n_finalized_blocks(n: usize, url: &str) {
 	use substrate_rpc_client::{ws_client, ChainApi};
 
 	let mut built_blocks = std::collections::HashSet::new();
-	let mut interval = tokio::time::interval(Duration::from_secs(3));
+	let mut interval = tokio::time::interval(Duration::from_secs(2));
```
amarts, line 182: leaving a note on this line as a learning for future.

After speaking to @smohan-dw, realized that this change was very crucial.

If this interval is more than block-time, then we 'miss' the events triggered.  This has to be lesser than block time.  (In this case, less than 3seconds)

test-utils/src/lib.rs
```diff
@@ -179,14 +179,14 @@ pub async fn wait_n_finalized_blocks(n: usize, url: &str) {
...
 	let rpc = ws_client(url).await.unwrap();
 
 	loop {
 		if let Ok(block) = ChainApi::<(), Hash, Header, ()>::finalized_head(&rpc).await {
 			built_blocks.insert(block);
 			if built_blocks.len() > n {
-				break
+				break;
```
amarts, line 189: Is this semicolon critical? I see another place where it got added?

amarts (approved):
With option changes, this PR also adds sudo pallet.

test-utils/src/lib.rs
roynrishingha: Explicit break and return do require a semicolon as they're parsed as statements.

test-utils/src/lib.rs
roynrishingha: Curious to know why `interval` is defined as mutable?

PR #184, opened 2023-06-28
Fix Github Action Release Error

roynrishingha:
Tested on a fork. Action prints the download link and we can finally download the asset. 
Currently it creates a release tag with the commit hash, and the current implementation runs on both `develop` and any branch name begins with `release` keyword.


amarts (approved):

amarts:
Looks like this is working, and lets remove `develop` branch from this.

PR #191, opened 2023-07-20
Build Docker Image using Buildx.

roynrishingha:
This workflow should should build the docker image. The docker image should support: x86-64 and ARM 64-bit architecture. QEMU hardware emulator is used.


roynrishingha:
Can be merged.

amarts (approved):

PR #199, opened 2023-08-01
Build score  pallet from scratch

rohibhan:


pallets/scoring/Cargo.toml
```diff
@@ -0,0 +1,66 @@
+[package]
+name = "pallet-scoring"
+description = "Manage ONDC Netwrok Scores."
```
amarts: Lets not 'restrict' this to ONDC Network.

```suggestion
description = "Manage Score of entities."
```

amarts (commented):
LGTM

pallets/scoring/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+use sp_runtime::Saturating;
+
+#[frame_support::pallet]
+pub mod pallet {
+	use super::*;
+	pub use cord_primitives::{curi::Ss58Identifier, CountOf, ScoreOf};
+	use cord_utilities::traits::CallSources;
+	use frame_support::pallet_prelude::*;
+	use frame_system::pallet_prelude::*;
+	use sp_runtime::traits::{IdentifyAccount, Saturating, Verify};
```
NiranjanAP: Getting a warning here - "warning: unused imports: `IdentifyAccount`, `Saturating`, `Verify`"

pallets/scoring/src/weights.rs
```diff
@@ -0,0 +1,24 @@
+#![allow(unused_parens)]
+#![allow(unused_imports)]
+
+// Dummy file
```
NiranjanAP: Dummy file?

rohibhan:
Need to add Registries for this pallet so test cases and benchmarking can change. Can you review and merge I will create another PR for Registry. Thanks

pallets/scoring/Cargo.toml
smohan-dw: or just say "Manage network scores"

pallets/scoring/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+	pub type CollectorIdentifierOf<T> = <T as frame_system::Config>::AccountId;
+
+	/// Hash of the Entry.
+	pub type EntryHashOf<T> = <T as frame_system::Config>::Hash;
+
+	pub type JournalIdentifierOf = IdentifierOf;
+	pub type RequestIdentifierOf = ScoreIdentifierOf;
+	pub type TransactionIdentifierOf = ScoreIdentifierOf;
+
+	pub type JournalDetailsOf<T> = JournalDetails<
```
smohan-dw: Align the storage names with the spec doc.

pallets/scoring/src/lib.rs
roynrishingha: Fixed it.

smohan-dw (dismissed):
LGTM

pallets/scoring/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+
+			Ok(())
+		}
+	}
+}
+
+impl<T: Config> Pallet<T> {
+	pub fn aggregate_score(entry: &RatingDetailsOf<T>) {
+		if let Some(mut aggregate) = <Scores<T>>::get(&entry.entity, &entry.rating_type) {
+			aggregate.count.saturating_inc();
```
smohan-dw: We need to check the rating type and act accordingly. A debit type should be added, and a credit type should reduce the existing score. For the credit type, we also need to ensure that the current value exceeds the credit transaction value.

smohan-dw: both count and value

pallets/scoring/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+				&journal.entry.entity,
+			);
+
+			Self::deposit_event(Event::JournalEntry {
+				identifier,
+				entity: journal.entry.entity.clone(),
+				author: author.clone(),
+			});
+
+			Self::aggregate_score(&journal.entry);
```
smohan-dw: This should be called before the DB entries, as the function will have a few error checks.

pallets/scoring/src/types.rs
```diff
@@ -0,0 +1,95 @@
...
+	/// transaction identifier for which the score is requested
+	pub tid: TransactionIdentfierOf,
+	///score collector identifier
+	pub collector: CollectorIdentifierOf,
+	///score requestor identifier
+	pub requestor: RequestorIdentifierOf,
+	/// score type
+	pub rating_type: RatingTypeOf,
+	///entity rating
+	pub rating: RatingOf,
```
smohan-dw: Missing entryType - should be an enum (debit, credit)

pallets/scoring/src/types.rs
```diff
@@ -0,0 +1,95 @@
...
+// (at your option) any later version.
+
+// CORD is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with CORD. If not, see <https://www.gnu.org/licenses/>.
+
```
smohan-dw, line 18: All type names should align with the names mentioned in the document - https://docs.google.com/document/d/1PWAOC9gWo0fL-crNrM-A1uXwzwNRY9jCMyc5L20_xRk/edit#heading=h.fnzqmubsnl03

runtime/src/lib.rs
```diff
@@ -960,6 +978,7 @@ construct_runtime! {
 		Schema: pallet_schema = 103,
 		Registry: pallet_registry = 104,
 		Stream: pallet_stream = 105,
+		Scoring: pallet_scoring = 163,
```
smohan-dw: 106

NiranjanAP:
Too many files in one PR - you should avoid this practice

runtime/src/lib.rs
```diff
@@ -978,7 +978,7 @@ construct_runtime! {
 		Schema: pallet_schema = 103,
 		Registry: pallet_registry = 104,
 		Stream: pallet_stream = 105,
-		Scoring: pallet_scoring = 163,
+		Scoring: pallet_scoring = 106,
 		DidNames: pallet_did_names = 106,
```
roynrishingha, line 983: Should both `Scoring` and `DidNames` have the same number `106`? What is this number btw ?

smohan-dw: It can't have the same number. This number represents the pallet index. A new pallet should take the next available number in the series. in this case scoring pallet number should be 107.

roynrishingha: Seems I have caught an early bug.

belyrs:
The link between RatingEntryDetails and ScoreEntry is missing, which is resulting in the inability to track the number of elements involved in the rating process.

amarts:
> The link between RatingEntryDetails and ScoreEntry is missing, which is resulting in the inability to track the number of elements involved in the rating process.

This is a good catch! Without this, we will be having discrepancy in what is expected from the pallet Vs what is implemented. Hence its a good practise to have cord.js changes in parallel.

amarts (changes requested):
To use `ScoreTypeOf` instead of `RatingOf` in extrinsic method.

PR #202, opened 2023-08-10
Makefile: A cluster of CORD Nodes

roynrishingha:
- Uses local chain specs with alice, bob, charlie, and dave accounts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,3 +105,33 @@ clean:
 	cargo clean
 
 .PHONY: build build-debug run-local help clean all $(TEST_CHECK_PALLETS)
+
+# Start a 4 node CORD network using local chain spec
+.PHONY: spinup spindown
+
+NODE_1_CMD := ./target/production/cord --base-path /tmp/data/alice --validator --chain local --alice --port 30333 --rpc-port 9933 --prometheus-port 9615 --node-key abe47f4e1065d4aa6fb0c1dd69a9a6b63c4551da63aad5f688976f77bd21622f --rpc-methods=Safe --rpc-cors all --no-hardware-benchmarks --state-pruning 100 --blocks-pruning 100 --offchain-worker never --prometheus-external --telemetry-url "wss://telemetry.cord.network/submit/ 0"
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -105,3 +105,33 @@ clean:
...
+
+
+spinup: build
+	@echo "Starting nodes in the background..."
+	@$(NODE_1_CMD) &
+	@$(NODE_2_CMD) &
+	@$(NODE_3_CMD) &
+	@$(NODE_4_CMD) &
+	@echo "Four CORD nodes (Alice, Bob, Charlie, Dave) have been successfully started."
+	@echo "See them in \033[0;34mtelemetry.cord.network\033[0m under Cord Spin tab."
```
amarts:
```suggestion
	@echo "See them in \033[0;34mtelemetry.cord.network\033[0m under Cord Spin tab. \nYou can also watch this network details in https://apps.cord.network/?rpc=ws://localhost:9933"
```

Makefile
```diff
@@ -105,3 +105,33 @@ clean:
...
+
+spindown:
+	@echo "Stopping all nodes..."
+	@pkill -f "cord --base-path /tmp/data/"
+	@echo "All nodes stopped."
+	@echo "Deleting /tmp/data/ directory..."
+	@rm -rf /tmp/data/
```
amarts: Just a word of caution! I feel `/tmp/data/` is too generic to be used by something else in developer laptop. Why don't we have it as `/tmp/cord-data/` or anything else which is bit more unique to CORD?

Makefile
```diff
@@ -105,3 +105,33 @@ clean:
...
+	@echo "/tmp/data/ directory deleted."
+	@echo "Bye Bye"
```
amarts: I feel 'Bye Bye' (twice) is not a great ending message, instead lets do something like 
"You can also connect with Dhiway to setup a private chain (sales@dhiway.com)"

"CORD team recommends having a separate chain in production, because 'local' chain uses the default keys, which are common." 

Something like above...

@smohan-dw @NiranjanAP @sankarshanmukhopadhyay

amarts (commented):
I guess this may be a better approach than docker-compose. Minor comments.

Makefile
sankarshanmukhopadhyay: > I feel 'Bye Bye' (twice) is not a great ending message, instead lets do something like "You can also connect with Dhiway to setup a private chain ([sales@dhiway.com](https://mail.google.com/mail/?view=cm&fs=1&tf=1&to=sales@dhiway.com))"
> 

If this is the code licensed under open source and available then co-mingling a commercial message might not be well received. If it is not, then something like "Commercial Support Services on CORD are offered by Dhiway <email>" might work

> "CORD team recommends having a separate chain in production, because 'local' chain uses the default keys, which are common."
> 

This is a very important message to convey

> Something like above...
> 
> @smohan-dw @NiranjanAP @sankarshanmukhopadhyay

roynrishingha:
@amarts  applied all your suggestions in the new commit.

amarts (approved):

PR #203, opened 2023-08-11
Makefile: Add persistent logging to all nodes and remove a few options from validator startup commands

roynrishingha:
Removed the flowing options:
  - `--no-hardware-benchmarks`
  - `--state-pruning 100`
  - `--blocks-pruning 100`
  - `--offchain-worker never`
 
Update `node-key` for all nodes.
Persist logs from all CORD nodes and print instructions to view the logs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,13 +109,13 @@ clean:
 # Start a 4 node CORD network using local chain spec
 .PHONY: spinup spindown
 
-NODE_1_CMD := ./target/production/cord --base-path /tmp/cord-data/alice --validator --chain local --alice --port 30333 --rpc-port 9933 --prometheus-port 9615 --node-key abe47f4e1065d4aa6fb0c1dd69a9a6b63c4551da63aad5f688976f77bd21622f --rpc-methods=Safe --rpc-cors all --no-hardware-benchmarks --state-pruning 100 --blocks-pruning 100 --offchain-worker never --prometheus-external --telemetry-url "wss://telemetry.cord.network/submit/ 0"
+NODE_1_CMD := ./target/production/cord --base-path /tmp/cord-data/alice --validator --chain local --alice --port 30333 --rpc-port 9933 --prometheus-port 9615 --node-key abe47f4e1065d4aa6fb0c1dd69a9a6b63c4551da63aad5f688976f77bd21622f --rpc-methods=Safe --rpc-cors all --prometheus-external --telemetry-url "wss://telemetry.cord.network/submit/ 0"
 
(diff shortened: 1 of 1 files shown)


amarts (dismissed):

Makefile
```diff
@@ -109,13 +109,13 @@ clean:
 # Start a 4 node CORD network using local chain spec
 .PHONY: spinup spindown
 
-NODE_1_CMD := ./target/production/cord --base-path /tmp/cord-data/alice --validator --chain local --alice --port 30333 --rpc-port 9933 --prometheus-port 9615 --node-key abe47f4e1065d4aa6fb0c1dd69a9a6b63c4551da63aad5f688976f77bd21622f --rpc-methods=Safe --rpc-cors all --no-hardware-benchmarks --state-pruning 100 --blocks-pruning 100 --offchain-worker never --prometheus-external --telemetry-url "wss://telemetry.cord.network/submit/ 0"
+NODE_1_CMD := ./target/production/cord --base-path /tmp/cord-data/alice --validator --chain local --alice --port 30333 --rpc-port 9933 --prometheus-port 9615 --node-key abe47f4e1065d4aa6fb0c1dd69a9a6b63c4551da63aad5f688976f77bd21622f --rpc-methods=Safe --rpc-cors all --prometheus-external --telemetry-url "wss://telemetry.cord.network/submit/ 0"
```
smohan-dw: For the local test network, the node key definition can be something like this
--node-key 0000000000000000000000000000000000000000000000000000000000000001
This gives a node id 12D3KooWEyoppNCUx8Yx66oV9fJnriXwCcXwDDUA2kj6vnc6iDE
We can remove the node key definitions from the remaining nodes, as they will all use the first node as the boot node.

roynrishingha: The node id for the suggested node key will be `12D3KooWEyoppNCUx8Yx66oV9fJnriXwCcXwDDUA2kj6vnc6iDEp`. A `p` is missing. But I don't understand why we don't need node-key for other 3 nodes!

amarts: node key is automatically created if not present. In case of this setup, only bootnode needs to have a recognisable node key. Other nodes can be random.

amarts (commented):
I checked the command and it is running fine. Meantime, I would suggest change in logging output.

we should log each node command to separate log file. Append ` 2>&1 > /tmp/cord-alice.log &` and similar for each node command, and mention it as output of make command to check those files to check log. That way, you can also use the same terminal for other activities.

roynrishingha:
> 

I used a different but more flexible approach for logs.

amarts (approved):

PR #204, opened 2023-08-11
Add Charlie as an authority to local chain spec

roynrishingha:
We had 2 authors: Alice and Bob.
Now adding Charlie to the genesis config of local chain spec.


smohan-dw (approved):
LGTM

amarts (approved):

PR #207, opened 2023-08-14
docker multiplatform image build

smohan-dw:
The multiplatform build needs to be executed on a dedicated machine. Github multiarch action hits OOM after 7 minutes. 

```docker buildx build --platform linux/amd64,linux/arm64 -f docker/Dockerfile -t dhiway/cord-multiarch:test . --push``` 

docker pull dhiway/cord-multiarch:test is the test image built using this dockerfile. 

Also, the final build image is changed to debian:bullseye-slim, as distroless arm64 missing a system library. We need to track this issue and change to distroless once it is fixed.

### CODE REVIEW

**Review by amarts — COMMENTED** [review]
LGTM

File: extras/README.md
Comment (amarts) [inline extras/README.md:1]: Looks like there is conflict here. May need a rebase.

File: .github/workflows/docker-multiarch.yml [hunk 1]
```diff
@@ -0,0 +1,62 @@
...
+
+    env:
+      REGISTRY: docker.io
+      USERNAME: dhiway
+      IMAGE_NAME: ${{ github.event.repository.name }}-multiarch-test
+      COMMIT_SHA: ${{ github.sha }}
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
```
roynrishingha, line 21: We need `QEMU` before `buildx`. Please add this:
```
      - name: Set up QEMU
        uses: docker/setup-qemu-action@v2
```

.github/workflows/docker-multiarch.yml
```diff
@@ -0,0 +1,62 @@
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
```
roynrishingha: Use this version 2
```
- name: Set up Buildx
  uses: docker/setup-buildx-action@v2
```

docker/Dockerfile
```diff
@@ -0,0 +1,50 @@
+# ------------------------------------------------------------------------------
+# Build Stage
+# ------------------------------------------------------------------------------
+
+FROM rust:1-bullseye as build
+
+LABEL maintainer="engineering@dhiway.com"
+
+ARG target
+ARG profile=release
```
roynrishingha, line 10: ARG PROFILE=production

Is not it better to make it all caps?

docker/Dockerfile
```diff
@@ -0,0 +1,50 @@
+# ------------------------------------------------------------------------------
+# Build Stage
+# ------------------------------------------------------------------------------
+
+FROM rust:1-bullseye as build
```
roynrishingha: FROM rust:1-bullseye as builder

docker/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+LABEL maintainer="engineering@dhiway.com"
+
+ARG target
+ARG profile=release
+
+WORKDIR /builder
```
roynrishingha: WORKDIR /build

docker/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+# Conditional installation of cross-compilation tools
+RUN if [ "$target" = "linux/arm64" ] || [ "$target" = "linux/arm64/v8" ]; then \
+  apt-get install -y gcc-aarch64-linux-gnu binutils-aarch64-linux-gnu g++-aarch64-linux-gnu && \
+  echo "export CC=aarch64-linux-gnu-gcc" >> /builder/dynenv && \
+  echo "export CXX=aarch64-linux-gnu-g++" >> /builder/dynenv; \
+  fi
+
+# Copy source tree
+COPY . /builder
```
roynrishingha: COPY . /build

docker/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+# Configure build environment and build
+RUN set -x && \
+  if [ -f /builder/dynenv ]; then . /builder/dynenv; fi && \
```
roynrishingha: if [ -f /build/dynenv ]; then . /build/dynenv; fi && \

docker/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+  cargo build --locked --profile $profile && \
+  mkdir -p build && \
+  mv target/$target/$profile/cord build/
+
+# ------------------------------------------------------------------------------
+# Final Stage
+# ------------------------------------------------------------------------------
+
+FROM debian:bullseye-slim
```
roynrishingha, line 39: Why not use alpine ?

docker/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+LABEL maintainer="engineering@dhiway.com"
+LABEL description="CORD Blockchain Node"
+
+COPY --from=build /builder/build/cord /cord
```
roynrishingha: Lets copy in `usr/local/bin` to directly use the binary.

docker/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+RUN ["/cord","--version"]
```
roynrishingha: RUN ["cord","--version"]

docker/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+USER 1000:1000
+EXPOSE 30333 9933 9944 9615
+
+ENTRYPOINT ["/cord"]
```
roynrishingha: ENTRYPOINT ["cord"]

docker/Dockerfile
smohan-dw: This is a temporary switch. Working with the distroless team to add the support we need

docker/Dockerfile
smohan-dw: not required as this is a temp switch

docker/Dockerfile
smohan-dw: not required as this is a temp switch

docker/Dockerfile
smohan-dw: not required as this is a temp switch

docker/Dockerfile
smohan-dw: It's good practice to keep the instructions in CAPS and arguments small.

docker/Dockerfile
smohan-dw: Yes, ci-linux is only available for AMD64. The image we are using is a multi-arch image from the Rust project.

docker/Dockerfile
smohan-dw: Any reason? I am using a build dir later. might cause confusion

.github/workflows/docker-multiarch.yml
smohan-dw: done, thanks for the catch

extras/README.md
smohan-dw: done

.github/workflows/docker-multiarch.yml
smohan-dw: done, thanks for the catch

docker/Dockerfile
roynrishingha: No specific reason. `/build` is mostly used, so I mentioned.

smohan-dw:
updated Dockerfile

docker/Dockerfile
smohan-dw: changed

.github/workflows/docker-multiarch.yml
```diff
@@ -0,0 +1,64 @@
+name: docker
+
+on:
+  push:
+    branches:
+      - develop
```
amarts, line 6: Can we keep the branch name as something unique (call it 'only-demo' or something) so it doesn't get triggered in github CI, but we use the logic on a builder in our infra to push it to DOCKER? With that, this is good to get in to repo.

amarts (approved):

.github/workflows/docker-multiarch.yml
smohan-dw: We can keep the docker actions as a reference. Post-merge, these actions (multiarch, docker) can be disabled at the repository level.

PR #212, opened 2023-08-19
[CI] Remove unused Github Actions

roynrishingha:
Docker Image release process has now changed, therefore, we don't need these Github Actions anymore.


amarts:
Why remove this? We want these files in the repo as a reference. We can 'disable' these workflow in CI.

roynrishingha:
@amarts closing this PR then.

PR #216, opened 2023-08-24
Fmt & update type in benchmark

siddharthteli12:
Ran `cargo fmt` & replaced const type in benchmark files to mock pallet type like - 

- `AUTHORITY_PROPOSALS `-> `T::MaxAuthorityProposals::get()` in authorship benchmark.
- `MAX_REGISTRY_SIZE` -> `T::MaxEncodedRegistryLength::get()` in registry benchmark.
- `MAX_SCHEMA_SIZE` -> `T::MaxEncodedSchemaLength::get()` in schema benchmark.

Also, not sure what `MAX_PAYLOAD_BYTE_LENGTH` const type used in stream & did benchmark refer to & what pallet type it can be replaced with.


amarts (commented):
LGTM

meantime, is there a guideline on using get() method instead of constant ?

siddharthteli12:
> LGTM
> 
> meantime, is there a guideline on using get() method instead of constant ?

Not really. But `get` will directly read mock type value. If mock is updated we need to manually update const value in benchmarking. Using const is just an overhead in most cases.

 Also, For eg, `const MaxAuthorityProposals` has already been defined in authorship mocks. It's better to utilise it rather than redeclaring it in benchmarking.

https://github.com/dhiway/cord/blob/86a025ec160fa25d7eda0bef2c769a7e37c4582a/pallets/authorship/src/mock.rs#L95C1-L97C2

smohan-dw (approved):
LGTM

amarts (approved):

PR #220, opened 2023-09-01
add rusty-hook as dev dependency

smohan-dw:


amarts (approved):
good. with this, before commit the formatting is fixed.

.editorconfig
```diff
@@ -1,10 +1,20 @@
 root = true
 
+
+[*]
+indent_style=space
+indent_size=2
```
roynrishingha, line 6: I think `rustfmt` uses 4 spaces.

.rusty-hook.toml
```diff
@@ -0,0 +1,5 @@
+[hooks]
+pre-commit = "cargo fmt --all -- --check"
```
roynrishingha, line 2: No certain how `rusty-hook` works. If it behaves quite similar to pre-commit hook, I think it's a nice idea to run clips as well. What do you think? Although it will take a significant time.

.editorconfig
smohan-dw: Right, we are keeping it 4 for *.rs files

.editorconfig
```diff
@@ -23,4 +33,10 @@ insert_final_newline=true
 indent_style=space
 indent_size=2
```
roynrishingha, line 34: As you mentioned we're keeping `indent_size=4`, but here it seems 2.

.editorconfig
```diff
@@ -1,10 +1,20 @@
...
+tab_width=2
+end_of_line=lf
+charset=utf-8
+trim_trailing_whitespace=true
+insert_final_newline = true
+
 [*.rs]
 indent_style=tab
 indent_size=tab
 tab_width=4
```
smohan-dw, line 16: @roynrishingha  indent_size for *.rs is 4, as it is mapped to tab_width

.editorconfig
smohan-dw: This is for shell files.

roynrishingha (approved):

PR #223, opened 2023-09-04
Core refactor - Governance, Membership & Payments 

smohan-dw:
- Authorship -> Network Membership
- Runtime upgrade pallet
- Pallet membership with network membership check
- Feeless transactions (transaction payment* removed)
- Pallet collective
- Authority Manager -> Authority Membership with network membership check
- Offence tracker linked to Authority Membership


node/client/Cargo.toml
```diff
@@ -2,19 +2,16 @@
...
 license = "GPL-3.0"
 homepage = "https://cord.network"
 repository = "https://github.com/dhiway/cord"
 
 [dependencies]
-async-trait = "0.1.57"
 futures = "0.3.21"
 frame-benchmarking = { git = "https://github.com/dhiway/substrate", branch = "release-v0.9.43" }
 frame-benchmarking-cli = { git = "https://github.com/dhiway/substrate", branch = "release-v0.9.43" }
-pallet-transaction-payment = { git = "https://github.com/dhiway/substrate", branch = "release-v0.9.43" }
```
roynrishingha, line 16: Question: Is transaction payment (Token) removed from the chain? For every signed transaction, an exchange of tokens (WAY) happens. Is it now removed?

node/service/src/chain_spec.rs
```diff
@@ -216,18 +216,26 @@ fn cord_local_genesis(
 	endowed_accounts: Option<Vec<AccountId>>,
 ) -> GenesisConfig {
 	let endowed_accounts: Vec<AccountId> = endowed_accounts.unwrap_or_else(testnet_accounts);
-
 	let num_endowed_accounts = endowed_accounts.len();
-	const ENDOWMENT: u128 = 50_000 * WAY;
+	const ENDOWMENT: u128 = 10_000 * WAY;
```
roynrishingha, line 220: Question: If the transaction payment is removed then why do we have/need `ENDOWMENT`?

pallets/collective/src/migrations/v4.rs
```diff
@@ -0,0 +1,153 @@
...
+/// Migrate the entire storage of this pallet to a new prefix.
+///
+/// This new prefix must be the same as the one set in construct_runtime. For
+/// safety, use `PalletInfo` to get it, as:
+/// `<Runtime as frame_system::Config>::PalletInfo::name::<CollectivePallet>`.
+///
+/// The migration will look into the storage version in order not to trigger a
+/// migration on an up to date storage. Thus the on chain storage version must
+/// be less than 4 in order to trigger the migration.
+pub fn migrate<T: frame_system::Config, P: GetStorageVersion + PalletInfoAccess, N: AsRef<str>>(
```
roynrishingha, line 41: Suggestion: For longer Trait bounds we can use `where` clause. It makes the code more readable.

pallets/network-membership/src/types.rs
```diff
@@ -0,0 +1,44 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with CORD. If not, see <https://www.gnu.org/licenses/>.
+
+use crate::*;
+use codec::{Decode, Encode};
+use frame_support::dispatch::DispatchClass;
+use scale_info::TypeInfo;
+#[cfg(feature = "std")]
+use serde::{Deserialize, Serialize};
```
roynrishingha, line 24: Question: When the runtime gets compiled to WASM, do we have access to std? WASM is no_std right? If so, then how the serialisation and deserialisation will work?  Here we have `serde` as a feature dependency.

pallets/network-membership/src/weights.rs
```diff
@@ -0,0 +1,129 @@
...
+// --extrinsic=*
+// --execution=Wasm
+// --wasm-execution=compiled
+// --heap-pages=4096
+// --output=./pallets/network-membership/src/weights.rs
+// --template=./.maintain/frame-weight-template.hbs
+
+#![cfg_attr(rustfmt, rustfmt_skip)]
+#![allow(unused_parens)]
+#![allow(unused_imports)]
```
roynrishingha, line 27: Suggestion: Before merging I think it's a good idea to remove unused imports.

pallets/offences/README.md
```diff
@@ -0,0 +1,10 @@
+# CORD offences pallet
+
```
roynrishingha, line 2: Suggestion: I think it is a good idea to write briefly what `offence` is.

pallets/offences/src/lib.rs
```diff
@@ -0,0 +1,264 @@
...
+	/// Other reports for the same report kinds.
+	concurrent_offenders: Vec<OffenceDetails<T::AccountId, T::IdentificationTuple>>,
+}
+
+/// An auxiliary struct for working with storage of indexes localized for a
+/// specific offence kind (specified by the `O` type parameter).
+///
+/// This struct is responsible for aggregating storage writes and the underlying
+/// storage should not accessed directly meanwhile.
+#[must_use = "The changes are not saved without called `save`"]
```
roynrishingha, line 232: Question: What does the `#[must_use = "something"]` macro do?

pallets/runtime-upgrade/Cargo.toml
```diff
@@ -0,0 +1,41 @@
+[package]
```
roynrishingha, line 1: Question: Runtime upgrade is done by uploading compressed wasm of runtime, right? Then why this pallet?

runtime/constants/src/lib.rs
```diff
@@ -44,7 +44,7 @@ pub mod time {
 	pub const MILLISECS_PER_BLOCK: Moment = 3000;
 	pub const SLOT_DURATION: Moment = MILLISECS_PER_BLOCK;
 	#[allow(clippy::identity_op)]
-	pub const EPOCH_DURATION: BlockNumber = prod_or_fast!(1 * HOURS, 2 * MINUTES);
+	pub const EPOCH_DURATION: BlockNumber = prod_or_fast!(2 * MINUTES, 1 * MINUTES);
```
roynrishingha, line 47: Question: What exactly `EPOCH_DURATION` does ?

roynrishingha (commented):
Have a few questions.

node/client/Cargo.toml
smohan-dw: Payment is removed. Balance is still required to enable the Genesis accounts. We can remove this requirement in a future PR.

smohan-dw: yes, it is removed.

node/service/src/chain_spec.rs
smohan-dw: Balance is required now to enable the genesis accounts. We can remove this in a future PR.

node/client/Cargo.toml
amarts: Yes, the 'transaction' payment is not present anymore, reducing number of events, and thus increasing the performance of the pallet's methods.

It will be truely based on authority now.

node/service/src/chain_spec.rs
amarts: Good question. 

Yes, transaction fee is removed, but not the 'Balance' in the chain. As we 'start' a network, we still need to define a set of entities with starting Balances in the chain. Hence the ENDOWMENT is required.

pallets/collective/src/benchmarking.rs
```diff
@@ -0,0 +1,652 @@
...
+			old_members.last().cloned(),
+			T::MaxMembers::get(),
+		)?;
+
+		// If there were any old members generate a bunch of proposals.
+		if m > 0 {
+			// Set a high threshold for proposals passing so that they stay around.
+			let threshold = m.max(2);
+			// Length of the proposals should be irrelevant to `set_members`.
+			let length = 100;
```
amarts, line 70: Should this `100` be defined properly?

pallets/collective/src/benchmarking.rs
```diff
@@ -0,0 +1,652 @@
...
+		assert_eq!(voting.ayes.len(), (m - 3) as usize);
+		assert_eq!(voting.nays.len(), 1);
+	}
+
+	close_early_disapproved {
+		// We choose 4 as a minimum so we always trigger a vote in the voting loop (`for j in ...`)
+		let m in 4 .. T::MaxMembers::get();
+		let p in 1 .. T::MaxProposals::get();
+
+		let bytes = 100;
```
amarts, line 294: same here.... why `100` ? why not 128, vs, why not 64? or similar.

pallets/collective/src/benchmarking.rs
```diff
@@ -0,0 +1,652 @@
...
+		assert_eq!(Collective::<T, I>::proposals().len(), (p - 1) as usize);
+		assert_last_event::<T, I>(Event::Executed { proposal_hash: last_hash, result: Err(DispatchError::BadOrigin) }.into());
+	}
+
+	close_disapproved {
+		// We choose 4 as a minimum so we always trigger a vote in the voting loop (`for j in ...`)
+		let m in 4 .. T::MaxMembers::get();
+		let p in 1 .. T::MaxProposals::get();
+
+		let bytes = 100;
```
amarts, line 454: one more place.

pallets/network-membership/src/types.rs
smohan-dw: We are using serde in a no_std context with derive feature for Serialize, Deserialize trait. check the cargo file

pallets/collective/src/migrations/v4.rs
```diff
@@ -0,0 +1,153 @@
...
+use frame_support::{
+	traits::{
+		Get, GetStorageVersion, PalletInfoAccess, StorageVersion,
+		STORAGE_VERSION_STORAGE_KEY_POSTFIX,
+	},
+	weights::Weight,
+};
+
+/// Migrate the entire storage of this pallet to a new prefix.
+///
```
amarts, line 33: @smohan-dw Considering we would be starting fresh with this network, do you foresee a 'migration' requirement? Not clear on why this would be used.

pallets/offences/src/lib.rs
smohan-dw: The must_use attribute is used to issue a diagnostic warning when a value is not "used". https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute

pallets/collective/src/tests.rs
```diff
@@ -0,0 +1,1418 @@
...
+
+		#[pallet::event]
+		#[pallet::generate_deposit(pub(super) fn deposit_event)]
+		pub enum Event<T: Config> {
+			ExternalProposed,
+		}
+	}
+}
+
+pub type MaxMembers = ConstU32<100>;
```
amarts, line 89: Ok, this is where having hardcoded value makes sense. as its just test.

pallets/runtime-upgrade/Cargo.toml
smohan-dw: This pallet sets the upgrade origin.

pallets/collective/src/migrations/v4.rs
smohan-dw: this is a one-time use code, which came as part of the fork. We can remove the migration part later.

runtime/constants/src/lib.rs
smohan-dw: sets the session time. We use sessions to rotate the validators/authorities.

pallets/network-membership/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+			Self::deposit_event(Event::MembershipRenewalRequested { member });
+
+			Ok(())
+		}
+
+		/// Revoke a membership. Only root or council orgin can perform this
+		/// action.
+		#[pallet::call_index(2)]
+		#[pallet::weight(<T as pallet::Config>::WeightInfo::revoke())]
+		pub fn revoke(origin: OriginFor<T>, member: CordAccountOf<T>) -> DispatchResult {
```
amarts, line 240: just a thought: should we keep an optional 'reason' string as an (256 bytes max) parameter. This needn't be stored anywhere, but be part of extrinsic so there is a reference (works good only if there are different reason for revoking a membership... like "inactive", "no show in voting for N months" etc.

pallets/network-membership/src/mock.rs
```diff
@@ -76,49 +78,46 @@ impl frame_system::Config for Test {
...
-	let keystore = sp_keystore::testing::MemoryKeystore::new();
-	#[cfg(feature = "runtime-benchmarks")]
-	ext.register_extension(sp_keystore::KeystoreExt(sp_std::sync::Arc::new(keystore)));
-	ext.execute_with(|| System::set_block_number(1));
-	ext
+pub(crate) fn default_gen_conf() -> NetworkMembershipConfig {
+	NetworkMembershipConfig {
+		members: btreemap![
+			 AccountId::new([11u8; 32]) => MemberData {
+				expire_on: 3,
```
amarts, line 119: May be a comment on why this `3` by default is a good to have.

amarts (commented):
Yet to review completely. Will do it in sometime. Wanted to post review comments of 30/80 files so far.

None of this is blocker, but an answer here or a new issue to implement it in the followup PR would be good.

pallets/collective/src/migrations/v4.rs
smohan-dw: agree. safe to remove the migration code.

amarts: Ok, lets keep it for now, and remove as a separate PR later.

pallets/network-membership/src/mock.rs
smohan-dw: this is to support test cases where the membership should expire at a certain block.

node/service/src/chain_spec.rs
roynrishingha: What exactly is the purpose of balance? How is it used internally on the chain?

pallets/collective/src/migrations/v4.rs
roynrishingha: Ok. In follow up PR, I would love to refactor then.

pallets/runtime-upgrade/Cargo.toml
roynrishingha: Upgrade origin is by default the Account used to run runtime upgrade, right ? I did not understand this. Can you plz go little deeper?

runtime/constants/src/lib.rs
roynrishingha: Private/public key pair changes in this interval? Why change the keys ??

pallets/collective/src/migrations/v4.rs
roynrishingha: Would love to raise a follow up PR removing migration code.

pallets/network-membership/src/lib.rs
smohan-dw: That's good to have. An enum-based approach would be nice compared to vec.

runtime/constants/src/lib.rs
smohan-dw: Session rotation provides a window to remove the offline authorities and add new authorities. No Key pair changes here.

amarts (approved):
LGTM (other than nitpick comments and few followup documentation, and minor corrections)

amarts:
Merging the PR with below todo lists

 - [ ] add more documentation on `offence` pallet (and what offence is)
 - [ ] add more information on runtime-upgrade pallet, and clarify some doubts
 - [ ] instead of having 100 has hardcoded value, can it be network configuration
 - [ ] No need of storage-migration logic in pallets as we will have a requirement to start from this version or higher, instead of migrating.

@roynrishingha @smohan-dw

smohan-dw:
> Merging the PR with below todo lists
> 
> * [ ]  add more documentation on `offence` pallet (and what offence is)

atm we are only tracking the disconnect offence. A disconnect offence is when an active validator goes offline. We have to implement the blacklist offence when a validator remains offline for a longer duration. 
 
> * [ ]  add more information on runtime-upgrade pallet, and clarify some doubts

This pallet provides a set_code origin for the technical committee to perform runtime upgrades. 

> * [ ]  instead of having 100 has hardcoded value, can it be network configuration

hardcoded values are used in tests/benchmarks. Pallet uses configurable parameters via runtime.

> @roynrishingha @smohan-dw

PR #227, opened 2023-09-12
Rohit/score pallet

rohibhan:
Score pallet credit and debit changes


pallets/scoring/src/lib.rs
```diff
@@ -0,0 +1,359 @@
...
+}
+
+impl<T: Config> Pallet<T> {
+	pub fn aggregate_score(entry: &RatingDetailsOf<T>) -> Result<(), pallet::Error<T>> {
+		if let Some(aggregate) = <Scores<T>>::get(&entry.entity, &entry.rating_type) {
+			let total_count = aggregate.count + entry.count;
+
+			let sum = match entry.entry_type {
+				RatingEntryType::Credit =>
+					Self::sum_average(aggregate.rating + entry.rating, total_count),
```
smohan-dw: Don't do average. We only need to keep the aggregate of counts and scores. The average function will be at the SDK level as the value received at the chain is base-10 encoded.

smohan-dw: also, use saturating_add and saturating_sub instead of +, - operators. (https://doc.rust-lang.org/std/intrinsics/fn.saturating_add.html#)

NiranjanAP: > Don't do average. We only need to keep the aggregate of counts and scores. The average function will be at the SDK level as the value received at the chain is base-10 encoded.

@adi-a11y   - FYI, the average calculation has to be performed in the SDK

pallets/scoring/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+			journal: RatingInputOf<T>,
+			authorization: AuthorizationIdOf,
+		) -> DispatchResult {
+			let author = <T as Config>::EnsureOrigin::ensure_origin(origin)?.subject();
+
+			let registry_id =
+				pallet_registry::Pallet::<T>::is_a_delegate(&authorization, author.clone(), None)
+					.map_err(<pallet_registry::Error<T>>::from)?;
+
+			ensure!(
```
smohan-dw, line 265: We can remove the MaxScoreValue check as the incoming entry is an aggregate.

pallets/scoring/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+	///Type for a Collector(Buyer) Identifier
+	pub type CollectorIdentifierOf<T> = <T as frame_system::Config>::AccountId;
+
+	pub type JournalIdentifierOf = IdentifierOf;
+	pub type RequestIdentifierOf<T> = ScoreIdentifierOf<T>;
+	pub type TransactionIdentifierOf<T> = ScoreIdentifierOf<T>;
+
+	pub type RatingDetailsOf<T> = RatingEntryDetails<
+		EntityIdentifierOf<T>,
+		RequestIdentifierOf<T>,
```
smohan-dw: RequestIdentifierOf & RequestorIdentifierOf are not part of the data model.

pallets/scoring/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+				(journal.entry.rating >= T::MinScoreValue::get() &&
+					journal.entry.rating <= T::MaxScoreValue::get()),
+				Error::<T>::InvalidRatingValue
+			);
+			ensure!(
+				!<JournalHashes<T>>::contains_key(journal.digest),
+				Error::<T>::DigestAlreadyAnchored
+			);
+
+			let identifier = Ss58Identifier::to_scoring_id(&(journal.digest).encode()[..])
```
smohan-dw: Identifiers from only journal hash will fail as duplicate entries are possible across days. We need to combine digest with TransactionIdentifierOf, RatingEntryTypeOf, RatingEntryOf, Requestor and Entity identifier to make it unique.

pallets/scoring/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+
+			Self::aggregate_score(&journal.entry)?;
+
+			<Journal<T>>::insert(
+				&identifier,
+				&journal.entry.rating_type,
+				RatingEntryOf::<T> {
+					entry: journal.entry.clone(),
+					digest: journal.digest,
+					created_at: block_number,
```
smohan-dw, line 303: should also capture transaction identifier.

pallets/schema/src/weights.rs
```diff
@@ -63,10 +63,10 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 		// Proof Size summary in bytes:
 		//  Measured:  `3`
 		//  Estimated: `18961`
-		// Minimum execution time: 25_673_000 picoseconds.
-		Weight::from_parts(26_208_134, 18961)
-			// Standard Error: 94
-			.saturating_add(Weight::from_parts(75_154, 0).saturating_mul(l.into()))
+		// Minimum execution time: 13_000_000 picoseconds.
```
smohan-dw, line 66: Discard weight file updates for the pallets unrelated to scoring from the commit.

amarts:
Hi @rohibhan. Consider rebasing this PR on top of `develop` branch.

Recommend you to take the latest code from this branch, and copy all new files on top of develop branch and then make required changes in runtime.

Do checkout #229 and a similar approach would be good.

pallets/scoring/src/lib.rs
rohibhan: property  `entry` which of type  `RatingEntryDetails`, already has `tid` as its field `tid: TransactionIdentfierOf`

PR #238, opened 2023-09-25
[CI] Fix static cord binary's dependency

roynrishingha:
This is a temporary fix by downgrading the runner-os. In future we need proper packaging solution to perfectly resolve this issue.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,22 +1,37 @@
+# A binary will be released when we push into any `release` branch
+# and/or we release it manually when we need.
+#
 name: release
 
 on:
   push:
     branches:
       - "release*"
+    tags:
+      - "v*.*.*"
+  workflow_dispatch:
+
+defaults:
+  run:
+    shell: bash
+
+env:
+  RUST_BACKTRACE: 1
+  CARGO_TERM_COLOR: always
 
 jobs:
   x86_64:
-    runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


roynrishingha:
@smohan-dw @amarts waiting review.

.github/workflows/release.yml
```diff
@@ -48,7 +63,7 @@ jobs:
           if [ -n "$BINARY_PATH" ]; then
             BRANCH_NAME=$(echo "${{ github.ref }}" | sed 's/refs\/heads\///')
             ARCHITECTURE=$(uname -m)
-            BINARY_NAME="cord-${BRANCH_NAME}-${ARCHITECTURE}-ubuntu-22.04"
+            BINARY_NAME="cord-${BRANCH_NAME}-${ARCHITECTURE}"
```
smohan-dw, line 66: Having different names for the binaries will break the tooling layer. Rather than building distribution-specific binaries, we should release distribution-specific packages. That workflow can be moved to another repo, as the package releases only need to happen after we tag a release candidate.

roynrishingha: This specific naming suggestion initially came from Amar. When a user downloads the binary our documentation force them to rename it directly to cord therefore tooling will keep working.

smohan-dw: That was long back. Since we are planning to move release activities to a separate repo, we don't need to make this change here. Safe to close this PR.

PR #239, opened 2023-09-27
Update Chain Storage Structure of Pallet Stream

roynrishingha:
- Add a new `StorageDoubleMap` named `UniqueProofs`
- Update this storage in `create`, `update` & `revoke` extrinsics.

### DIFF
--- a/pallets/stream/src/lib.rs
+++ b/pallets/stream/src/lib.rs
@@ -47,7 +47,10 @@ pub use crate::{pallet::*, types::*, weights::WeightInfo};
 pub mod pallet {
 	use super::*;
 	use cord_utilities::traits::CallSources;
-	use frame_support::pallet_prelude::*;
+	use frame_support::{
+		pallet_prelude::{StorageDoubleMap, *},
+		Blake2_128Concat, Twox64Concat,
+	};
 	use frame_system::pallet_prelude::*;
 	use sp_runtime::traits::Hash;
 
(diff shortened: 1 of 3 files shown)


pallets/stream/src/lib.rs
```diff
@@ -328,10 +356,31 @@ pub mod pallet {
 				ensure!(stream_details.registry == registry_id, Error::<T>::UnauthorizedOperation);
 			}
 
+			let registry_id =
```
smohan-dw: Identifier creation is only required at the stream creation level. We don't need it here.

pallets/stream/src/lib.rs
```diff
@@ -328,10 +356,31 @@ pub mod pallet {
...
 			<Streams<T>>::insert(
 				&stream_id,
+				StreamEntryOf::<T> {
+					digest: stream_digest,
+					creator: updater.clone(),
+					..stream_details.clone()
+				},
+			);
+
+			<UniqueProofs<T>>::insert(
```
smohan-dw, line 374: Before we add another proof, we need to revoke the previous proof.

pallets/stream/src/lib.rs
```diff
@@ -389,9 +438,37 @@ pub mod pallet {
 				ensure!(stream_details.registry == registry_id, Error::<T>::UnauthorizedOperation);
 			}
 
+			let registry_id =
```
smohan-dw: not required here.

smohan-dw (changes requested):

pallets/stream/src/lib.rs
```diff
@@ -492,6 +575,8 @@ pub mod pallet {
 
 			<Streams<T>>::take(&stream_id);
 
+			<Commits<T>>::take(&stream_id);
```
smohan-dw: This is not correct. This will erase the complete history. We need to read from the commit and iterate the result to remove the entries from unique proofs.

pallets/stream/src/lib.rs
```diff
@@ -490,6 +573,8 @@ pub mod pallet {
 				ensure!(stream_details.registry == registry_id, Error::<T>::UnauthorizedOperation);
 			}
 
+			<UniqueProofs<T>>::remove(&stream_id, stream_details.digest);
```
smohan-dw: This needs to be a loop. An identifier can have multiple proof entries in Uniqueproofs. We need to make sure all of them are removed. The list of digest belong to an identifier can be fetched from commits

roynrishingha:
@smohan-dw @amarts closing this PR as this will not be merged.

PR #241, opened 2023-10-02
pallet_stream: double map to provide latest state for a given 'identifer:hash'

amarts:
This is a sample on adding a doublemap to get the status of given hash / digest.

### DIFF
--- a/pallets/stream/src/lib.rs
+++ b/pallets/stream/src/lib.rs
@@ -78,8 +78,8 @@ pub mod pallet {
 	pub type StreamEntryOf<T> =
 		StreamEntry<StreamDigestOf<T>, StreamCreatorIdOf<T>, SchemaIdOf, RegistryIdOf>;
 	/// Type for the stream digest entity
-	pub type StreamAttestationEntryOf<T> =
-		StreamAttestationEntry<StreamCreatorIdOf<T>, StreamRevokedByIdOf<T>, StatusOf>;
+	pub type AttestationDetailsOf<T> =
+		AttestationDetails<StreamCreatorIdOf<T>, StreamRevokedByIdOf<T>, StatusOf>;
 	/// Type for the stream commits
 	pub type StreamCommitsOf<T> = StreamCommit<
 		StreamCommitActionOf,
@@ -149,7 +149,7 @@ pub mod pallet {
 		StreamIdOf,
 		Blake2_128Concat,
 		StreamDigestOf<T>,
-		StreamAttestationEntryOf<T>,
+		AttestationDetailsOf<T>,
 		OptionQuery,
 	>;
 
@@ -295,7 +295,7 @@ pub mod pallet {
 			<Attestations<T>>::insert(
 				&identifier,
 				stream_digest,
-				StreamAttestationEntryOf::<T> {
+				AttestationDetailsOf::<T> {
 					creator: creator.clone(),
 					revoked_by: None,
 					revoked: false,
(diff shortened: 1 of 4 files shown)


amarts:
wanted to dust-off my coding on cord. @smohan-dw I thought having just the status is good enough in this query, as identifier's query would be having all the other details. This will make verification proper IMO.

Yet to test, but meantime wanted feedback if I am very much off the track.

Updates #239

pallets/stream/src/lib.rs
```diff
@@ -267,12 +281,14 @@ pub mod pallet {
...
 					digest: stream_digest,
 					creator: creator.clone(),
-					schema: schema_id,
+					schema: schema_id.clone(),
 					registry: registry_id,
 					revoked: false,
 				},
 			);
 
+			<UniqueStreams<T>>::insert(&identifier, stream_digest, false);
```
smohan-dw: +1. this can become the revocation registry as well. We can think of removing the revoke field from the StreamEntry and make this structure the state proof.

roynrishingha (commented):
@amarts is not this PR same as my PR?

amarts:
Similar, not same. 

Changes:

* Structure of what is stored.
* No extra fetch for digest, and reuse the existing.
* Handle all cases for keeping the <identity:hash> key's status up-to-date.

pallets/stream/src/lib.rs
amarts: True. Just that we have to make it clear, (or do 2 queries from SDK) that one is state check, and another is identifier latest details check.

amarts: moved 'creator' and 'revoked_by' also into the structure.

amarts:
Few updates:

* Name the storage as 'Attestations'.
* Add 'creator' and 'revoked_by' fields into attestation, as they can be different accounts which can be performing operations on same identifiers.
* Removed 'revoked' flag from Identifier (which means, the query should change in cord.js verification methods.
* stream::remove(), removes all entries in the given identifier.

pallets/stream/src/lib.rs
```diff
@@ -135,6 +140,20 @@ pub mod pallet {
...
+	/// It maps from a stream identifier and hash to its details.
+	#[pallet::storage]
+	#[pallet::getter(fn attestations)]
+	pub type Attestations<T> = StorageDoubleMap<
+		_,
+		Twox64Concat,
+		StreamIdOf,
+		Blake2_128Concat,
+		StreamDigestOf<T>,
+		StreamAttestationEntryOf<T>,
```
smohan-dw: This can be StreamEnty (or rename it to AttestationDetails). With that, We can change Streams to StreamIdentifier, which maps an identifier to the latest digest.

amarts: updated this. 

Other storage update IMO can come in as separate PR. So, that review will be cleaner.

roynrishingha (dismissed):

roynrishingha (approved):

smohan-dw:
LGTM. Please hold the merge for a few hours. Almost completing the SDK update

smohan-dw (commented):
LGTM. We should get the SDK updated before the merge. SDK changes are required as the storage parameters are changed.

amarts:
> LGTM. Please hold the merge for a few hours. Almost completing the SDK update

@smohan-dw done - https://github.com/dhiway/cord.js/pull/132

PR #242, opened 2023-10-04
SDK update - v1.1.0

smohan-dw:
- Update to Substrate SDK v1.1.0
- Switch to Polkadot SDK repo
- Update to the latest Benchmark Syntax
- Pallet Renames
  - pallet-did-names -> pallet-did-name
  - pallet-scoring -> pallet-score


node/src/service.rs
```diff
@@ -328,6 +328,7 @@ pub fn new_partial(
 		client.clone(),
 	);
 
+	#[allow(clippy::redundant_clone)]
 	let (grandpa_block_import, grandpa_link) = sc_consensus_grandpa::block_import(
 		client.clone(),
```
roynrishingha, line 333: Can we not remove the clone and borrow?

roynrishingha (approved):

node/src/service.rs
smohan-dw: can try it as a separate activity. might have some additional dependencies

amarts (approved):

PR #245, opened 2023-10-09
Pallet: change name of 'Stream' to 'Statement'

amarts:
Along with changing the name of the pallet. Few other key things are altered in behavior in this PR.

* create() method takes Vec array of digests as argument, so it can create 'multiple' statement identifiers in one create call. (Note, author should be same account, and registry should be same for this).

* Digest to Identifier mapping is changed to `DoubleStorageMap`, considers the registryIdentifier as second key.

* `api.query.stream.commits()` -> `api.query.statement.activities()`


pallets/statement/src/lib.rs
```diff
@@ -0,0 +1,748 @@
...
+				)
+				.map_err(<pallet_registry::Error<T>>::from)?;
+			}
+
+			// now check the digests array
+			let digests_len = statement_digests.len();
+			ensure!(digests_len <= T::MaxDigestLength::get(), Error::<T>::MaxDigestLengthExceeded);
+
+			let mut event_array: Vec<StatementDigestOf<T>> = vec![];
+			for (_, statement_digest) in statement_digests.into_iter().enumerate() {
```
smohan-dw: This can be just `for statement_digest in statement_digests` since you are not using the index

pallets/statement/src/lib.rs
```diff
@@ -0,0 +1,748 @@
...
+						&registry_id.clone().encode()[..],
+						&creator.clone().encode()[..],
+					]
+					.concat()[..],
+				);
+
+				let identifier = Ss58Identifier::to_statement_id(&(id_digest).encode()[..])
+					.map_err(|_| Error::<T>::InvalidIdentifierLength)?;
+
+				if <Statements<T>>::contains_key(&identifier) {
```
smohan-dw, line 307: This will fail for all new identifiers as the storage is populated later.

amarts:
Merging everything to `dev-0.9.0` branch for now. Please merge future enhancement there.

PR #247, opened 2023-10-13
[FEAT] Bootstrap Configuration Support 

smohan-dw:
The PR brings the capability to generate a new chain spec using a TOML configuration file. The TOML configuration file can be generated by running the `bootstrap.sh` script. More documentation to follow.


amarts (approved):

roynrishingha (changes requested):
CI fails on Clippy. It is a better practice to fix it before merge.

roynrishingha (approved):

PR #249, opened 2023-11-04
Refactor - Identifiers & Statements

smohan-dw:
Identifiers are moved into primitives to ensure multiple pallets can use them as a global store representing identifier history. Statement is refactored to support both individual statement creation and batch creation.


amarts:
The build is passing. And I will work on the clippy and test part. Merging it to dev-0.9.0 branch so we can collaborate faster.

PR #254, opened 2023-11-29
experimental asset benchmarking

smohan-dw:


amarts (approved):
A rebase on latest of dev-0.9.0 would be great.

amarts:
one question is, do we need number accounting here? 

Like if I need to create an asset of 100000, should I call 100000 transaction to chain?

amarts (approved):

PR #255, opened 2023-11-30
Multiple changes for release v0.9.0

amarts:
# New features

* Fees: 
  - no more fees involved in transaction.
  - the events are removed, which increases the performance

* membership: 
  - Anyone (Any Account) who would 'write' (ie, perform extrinsics call) needs to have membership registration. This is approved by the council. 

* network-membership: 
  - this is managing the validator node accounts. This enabled us to remove the staking based logic, and now the logic is truly proof of authority. Again, the addition and removal is handled by council calls.

* chain-space:
  - this pallet now handles transaction delegation and limit of number of transactions on the chain authorization.
  - Think of this as a namespace, where most of the management logic would be implemented (like delegation, capacity etc)

* score:
  - allows registering and amends to the 'rating', and manages the aggregate score

* statement:
   - Allows register,  revoke, restore, update, and remove of 'statements' (combination of identifier and digest)
   - Allows presentation with its 'digest' to be registered.
   - `schemaUri` is  optional parameter, thus allowing different VCs in same `chain-space`.
 
* schema:
   - made as an independent call, inside of chain-space.
   
* asset:
   - Early beta of asset management pallet.


smohan-dw (approved):
LGTM

PR #261, opened 2023-12-23
update to sdk-v1.5.0

smohan-dw:


amarts:
Few questions. Should we do this subsecond block time as the default? What impact would that have on network requirements in nodes?

I see some node code reorg and tests added. Any particular top level changes to publish in release notes?

smohan-dw:
> Few questions. Should we do this subsecond block time as the default? What impact would that have on network requirements in nodes?
The updates are not for production, only to test the behaviour with multiple blocktimes. 
> 
> I see some node code reorg and tests added. Any particular top level changes to publish in release notes?
Nothing more than SDK updates, code reorg and tests in this PR.

smohan-dw:
looks like we need to make a few changes to the workflow file.

amarts:
```
error: failed to run custom build command for `cord-runtime v0.9.1 (/home/runner/work/cord/cord/runtime)`
note: To improve backtraces for build dependencies, set the CARGO_PROFILE_DEV_BUILD_OVERRIDE_DEBUG=true environment variable to enable debug information generation.

Caused by:
  process didn't exit successfully: `/home/runner/work/cord/cord/target/debug/build/cord-runtime-a02d8401b1f833b2/build-script-build` (exit status: 1)
  --- stderr
  Cannot compile the WASM runtime: no standard library sources found!
  You can install them with `rustup component add rust-src` if you're using `rustup`.
warning: build failed, waiting for other jobs to finish...
Error: Process completed with exit code 101.

```

Working fine on local machine.

amarts:
Updated the github workflow file by removing the caching section. Will check if the tests pass now.

amarts (approved):

PR #264, opened 2024-01-02
Update Benchmarks

smohan-dw:


smohan-dw:
Need to run this again on a machine with an NvEME disk and a CPU with better single-thread performance.

[-] Failed the machine benchmark:\n2024-01-02 18:40:21 Running machine benchmarks...
2024-01-02 18:40:47
+----------+----------------+-------------+-------------+-------------------+
| Category | Function       | Score       | Minimum     | Result            |
+===========================================================================+
| CPU      | BLAKE2-256     | 863.60 MiBs | 783.27 MiBs | ✅ Pass (110.3 %) |
|----------+----------------+-------------+-------------+-------------------|
| CPU      | SR25519-Verify | 402.96 KiBs | 560.67 KiBs | ❌ Fail ( 71.9 %) |
|----------+----------------+-------------+-------------+-------------------|
| Memory   | Copy           | 4.52 GiBs   | 11.49 GiBs  | ❌ Fail ( 39.4 %) |
|----------+----------------+-------------+-------------+-------------------|
| Disk     | Seq Write      | 240.02 MiBs | 950.00 MiBs | ❌ Fail ( 25.3 %) |
|----------+----------------+-------------+-------------+-------------------|
| Disk     | Rnd Write      | 119.94 MiBs | 420.00 MiBs | ❌ Fail ( 28.6 %) |
+----------+----------------+-------------+-------------+-------------------+
From 5 benchmarks in total, 1 passed and 4 failed (10% fault tolerance).

pallets/schema/src/weights.rs
```diff
@@ -48,44 +47,56 @@
...
-		// Minimum execution time: 25_673_000 picoseconds.
-		Weight::from_parts(26_208_134, 18961)
-			// Standard Error: 94
-			.saturating_add(Weight::from_parts(75_154, 0).saturating_mul(l.into()))
-			.saturating_add(RocksDbWeight::get().reads(1_u64))
-			.saturating_add(RocksDbWeight::get().writes(1_u64))
+		//  Measured:  `661`
+		//  Estimated: `19007`
+		// Minimum execution time: 55_169_000 picoseconds.
+		Weight::from_parts(57_651_421, 19007)
```
amarts: This is the one place where we are passing variable length argument. Should the weight change based on that?

pallets/statement/src/lib.rs
```diff
@@ -766,7 +766,7 @@ pub mod pallet {
 		///   the number of entries
 		/// removed.
 		#[pallet::call_index(4)]
-		#[pallet::weight(<T as pallet::Config>::WeightInfo::remove( ))]
+		#[pallet::weight(<T as pallet::Config>::WeightInfo::remove(T::MaxRemoveEntries::get() as u32))]
```
amarts, line 769: would this require a version change?

runtime/constants/src/weights/block_weights.rs
```diff
@@ -46,17 +44,17 @@ parameter_types! {
 	/// Calculated by multiplying the *Average* with `1.0` and adding `0`.
 	///
 	/// Stats nanoseconds:
-	///   Min, Max: 218_200, 244_112
-	///   Average:  223_580
-	///   Median:   221_485
-	///   Std-Dev:  5572.62
+	///   Min, Max: 280_638, 577_472
```
amarts: Compared to earlier a lot of diff between min and max. Is this alright?

amarts (dismissed):
LGTM

Few question. The benchmark runs, should we do that in CI machines? (and thus make this CI fail depending on the type of machine we get in github?) or is there an option to run it specifically on some setup with docker, so we can ask output of this command to be sent to us before starting a validator node ?

runtime/constants/src/weights/block_weights.rs
smohan-dw: Average has not moved up much. Couple of reasons, increased blocktime and WASM

amarts (approved):

amarts:
For those who are referring, the hardware tests are not part of this PR, but will be added back as a separate PR, so we can discuss more on it with focus.

PR #265, opened 2024-01-04
runtime weight and time improvements

smohan-dw:
❄️  Transaction Benchmarking (Baseline)
  🔖  Preparing 5000 transactions took 0.167s

  🎁  Anchoring 5000 individual transactions took 6.052s
  🙌  Block TPS (individual transactions) - 826

  🎁  Anchoring a batch of 5000 transactions took 0.047s
  🙌  Block TPS (batch transactions) - 106383

### DIFF
--- a/runtime/constants/src/lib.rs
+++ b/runtime/constants/src/lib.rs
@@ -40,11 +40,23 @@ pub mod currency {
 
 /// Time and blocks.
 pub mod time {
-	use cord_primitives::{prod_or_fast, BlockNumber, Moment};
+	use cord_primitives::{BlockNumber, Moment};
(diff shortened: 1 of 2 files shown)


amarts (approved):
1 sec block time, 
2 min epoch,
50% saturation instead of 80% (this is something I am not clear yet on the reasoning and impact).

smohan-dw:
> 1 sec block time,
> 2 min epoch,
> 50% saturation instead of 80% (this is something I am not clear yet on the reasoning and impact).

It's a 3-second block time, with 1 second for computation and remaining for propagation, validation, and voting. Kept the block saturation at 50% to start with as a safe-guard for weight calculations. This number can be increased later as we improve the benchmarks and hardware recommendations.

PR #266, opened 2024-01-11
fix: pallet_asset: use did origin check, instead of account

amarts:


amarts:
@smohan-dw check if these are in right track. Anything major I missed?

Also I was thinking if transfers should have the authorization check at all. As long as signature check passes, it should go through IMO to keep it much faster than create/issue

smohan-dw (dismissed):
LGTM. We should remove the commented code before merging.

amarts:
This can wait till we get the corresponding changes in SDK (@cord.network/asset).

Wanted to have it here as a reference.

smohan-dw (approved):

PR #269, opened 2024-01-18
fix: add didAuthorization to pallet/asset

amarts:


amarts:
This is a fix on top of #266

smohan-dw (approved):

PR #275, opened 2024-02-01
fix: bootstrap.sh: allow re-use-of-nodekeys

amarts:
assuming an admin/script would re-use the directory for same type of network (`sprint/` for sprintnet, `spark/` for sparknet etc), this minor change allows 're-use' of node key, which helps in automating the bootnode option encoding in scripts if someone is automating.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -88,8 +88,9 @@ generate_address() {
 }
 
 generate_node_key() {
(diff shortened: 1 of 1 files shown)


smohan-dw (approved):

amarts:
surprised to see clippy fail, as this is completely not script run. For now, will merge this without re-triggering the run.

Issue #268, opened 2024-01-11
pallet_asset: implement a restore() call


amarts:
After #267 is implemented, there would be a need to restore the revoked asset / instance.

This may help in a 'suspension' of asset to see if the particular transfer query would come on the revoked asset.

amarts:
This issue is no more required as we implemented 'status-change' instead of revoke - restore. now same call can be used for all state change.

PR #277, opened 2024-02-04
feat: pallet/asset: implement status_change() call

### LINKED ISSUE #267: pallet_asset: implement revoke() call
refer to revoke call in statement, and bring similar functionality in asset.

Expectation: revoke an asset or an instance (if its already active, and owner is matching).

amarts:
Fixes #267

### DIFF
--- a/pallets/asset/src/lib.rs
+++ b/pallets/asset/src/lib.rs
@@ -168,7 +168,11 @@ pub mod pallet {
 		},
 		/// An asset entry has been revoked.
 		/// \[asset entry identifier, optional instance identifier\]
-		Revoke { identifier: AssetIdOf, instance: Option<AssetInstanceIdOf> },
+		StatusChange {
+			identifier: AssetIdOf,
+			instance: Option<AssetInstanceIdOf>,
+			status: AssetStatusOf,
+		},
 	}
 
 	#[pallet::error]
@@ -199,6 +203,8 @@ pub mod pallet {
 		DistributionLimitExceeded,
 		/// asset instance not found
 		AssetInstanceNotFound,
+		/// Asset is in same status as asked for
+		AssetInSameState,
 	}
 
 	#[pallet::call]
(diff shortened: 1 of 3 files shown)


pallets/asset/src/lib.rs
```diff
@@ -383,6 +386,59 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		#[pallet::call_index(3)]
+		#[pallet::weight({0})]
+		pub fn revoke(
```
smohan-dw: Revoke in asset context is a status change according to the AssetStatusOf enum. If the asset status is inactive or expired, no more issuance should happen. We need to check the expected behaviour of transfers if the underlying asset is not active.

pallets/asset/src/lib.rs
```diff
@@ -383,6 +386,59 @@ pub mod pallet {
...
+			origin: OriginFor<T>,
+			asset_id: AssetIdOf,
+			instance_id: Option<AssetInstanceIdOf>,
```
smohan-dw, line 401: We need to implement the status change behaviour for assets and instances separately.

amarts:
As per one of the comment, instead of focusing on individual status of the asset/instance, one can keep `status_change()` itself as the key call, which can handle both revoke, restore. 

fixes: #267, #268

amarts:
@smohan-dw does this makes sense now?

@adi-a11y the companion PR in cord.js would change too.

pallets/asset/src/lib.rs
```diff
@@ -166,6 +166,13 @@ pub mod pallet {
 			from: AssetCreatorOf<T>,
 			to: AssetCreatorOf<T>,
 		},
+		/// An asset entry has been revoked.
```
amarts: change the comment to reflect the new changes

amarts:
@smohan-dw as per the discussion, for now, only 3 states are good, and this `status_change()` call handles asset status change or instance status change depending on the argument. 

Good to go IMO. 

I know this needs a version change in runtime, when deployed. But I am not sure if that should be with this PR, or a separate PR before a release (mentioning all changes in one shot).

smohan-dw (dismissed):

pallets/asset/src/lib.rs
amarts: @smohan-dw looks like the comment is bit older than the review. Currently, the status change takes instance status change if instance is sent. Asset state is not changed. If instance is None, then asset status change happens.

pallets/asset/src/lib.rs
amarts: The logic of transfer and issuance only checks for 'ACTIVE' already in code, so any other status change would result in stopping of transfer and issuance from happening.

amarts:
Merging this as @smohan-dw's approval was present for commit without the version change. With this, we will be comfortable to update the existing chain without any major issues.

PR #324, opened 2024-02-09
feat: chain-space: implement a subspace logic

amarts:
chain-space pallet is designed to handle delegate management and also number of transactions (approved from the council). This way, we always endup with a proper authorization as the way to write to chain, and manage multiple usecases / projects with the same DID for an entity. 

This allows entities (legal-person (ie, humans, orgs, councils, etc)) to keep same Identity on chain and manage multiple usecases, thus solving ACL related issues on chain.

One of the limitation of having council approve all the chain-space count is, it becomes un-manageable. With this subspace concept, the governance council can only approve each nodes (or possible network-members) with a space, and give say N Million transaction capability on the chain, and later, each of the creator of these spaces can continue to carve out usecase specific spaces from the same.

### DIFF
--- a/pallets/chain-space/src/benchmarking.rs
+++ b/pallets/chain-space/src/benchmarking.rs
@@ -386,6 +386,40 @@ benchmarks! {
 			 assert_last_event::<T>(Event::ApprovalRestore { space: space_id }.into());
 		 }
 
+		subspace_create {
+			 let caller: T::AccountId = account("caller", 0, SEED);
+			 let did: T::SpaceCreatorId = account("did", 0, SEED);
+			 let space = [2u8; 256].to_vec();
+			 let subspace = [5u8; 256].to_vec();
+			 let capacity = 50u64;
+
+			 let space_digest = <T as frame_system::Config>::Hashing::hash(&space.encode()[..]);
+			 let subspace_digest = <T as frame_system::Config>::Hashing::hash(&subspace.encode()[..]);
+			 let id_digest = <T as frame_system::Config>::Hashing::hash(
+				 &[&space_digest.encode()[..], &did.encode()[..]].concat()[..],
+			 );
+			 let space_id: SpaceIdOf = generate_space_id::<T>(&id_digest);
+
+			 let sub_id_digest = <T as frame_system::Config>::Hashing::hash(
+				 &[&subspace_digest.encode()[..], &did.encode()[..]].concat()[..],
+			 );
+			 let subspace_id: SpaceIdOf = generate_space_id::<T>(&sub_id_digest);
+
+			 let auth_id_digest = <T as frame_system::Config>::Hashing::hash(
+				 &[&subspace_id.encode()[..], &did.encode()[..]].concat()[..],
+			 );
+			 let authorization_id: AuthorizationIdOf = generate_authorization_id::<T>(&auth_id_digest);
+
+			 let origin =  <T as Config>::EnsureOrigin::generate_origin(caller, did);
+
+			 Pallet::<T>::create(origin.clone(), space_digest )?;
+			 Pallet::<T>::approve(RawOrigin::Root.into(), space_id.clone(), capacity )?;
+
+		 }: _<T::RuntimeOrigin>(origin.clone(), subspace_digest, 10u64, space_id.clone())
+		 verify {
+			 assert_last_event::<T>(Event::Create { space: subspace_id, creator: origin, authorization: authorization_id }.into());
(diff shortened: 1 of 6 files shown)


amarts:
a step towards achieving https://github.com/cord-network/community/issues/3

pallets/chain-space/src/lib.rs
```diff
@@ -915,6 +916,129 @@ pub mod pallet {
...
+		///
+		/// # Errors
+		/// - `InvalidIdentifierLength`: If the generated identifiers for the space or authorization
+		///   are of invalid length.
+		/// - `SpaceAlreadyAnchored`: If the space identifier is already in use.
+		/// - `SpaceDelegatesLimitExceeded`: If the space exceeds the limit of allowed delegates.
+		///
+		/// # Events
+		/// - `Create`: Emitted when a new space is successfully created. It includes the space
+		///   identifier, the creator's identifier, and the authorization ID.
```
amarts, line 993: Note: The comments needs to be updated to reflect the actual function.

Also, the event to be raised, can be 'SubSpaceCreate' or 'Create' itself. Need to discuss.

pallets/chain-space/src/lib.rs
```diff
@@ -915,6 +916,129 @@ pub mod pallet {
...
+			space_code: SpaceCodeOf<T>,
+			count: u64,
+			space_id: SpaceIdOf,
+		) -> DispatchResult {
+			let creator = <T as Config>::EnsureOrigin::ensure_origin(origin)?.subject();
+
+			let space_details = Spaces::<T>::get(&space_id).ok_or(Error::<T>::SpaceNotFound)?;
+			ensure!(!space_details.archive, Error::<T>::ArchivedSpace);
+			ensure!(space_details.approved, Error::<T>::SpaceNotApproved);
+
```
amarts, line 1008: Need to add a test for creator check, which should match with the parent space creator.

pallets/chain-space/src/lib.rs
```diff
@@ -915,6 +967,138 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		/// Creates a new space with a unique identifier based on the provided
```
vatsa287, line 957: Are the comments auto-generated or defined?

amarts: Comments are written. There is a tool / extension in few editors to write comments, you can use them.

pallets/chain-space/src/benchmarking.rs
```diff
@@ -386,6 +386,40 @@ benchmarks! {
...
+			 let authorization_id: AuthorizationIdOf = generate_authorization_id::<T>(&auth_id_digest);
+
+			 let origin =  <T as Config>::EnsureOrigin::generate_origin(caller, did);
+
+			 Pallet::<T>::create(origin.clone(), space_digest )?;
+			 Pallet::<T>::approve(RawOrigin::Root.into(), space_id.clone(), capacity )?;
+
+		 }: _<T::RuntimeOrigin>(origin.clone(), subspace_digest, 10u64, space_id.clone())
+		 verify {
+			 assert_last_event::<T>(Event::Create { space: subspace_id, creator: origin, authorization: authorization_id }.into());
```
vatsa287: In `assert_last_event`, I think `creator` should be `did.clone()`

pallets/chain-space/src/benchmarking.rs
```diff
@@ -386,6 +386,40 @@ benchmarks! {
...
+				 &[&subspace_digest.encode()[..], &did.encode()[..]].concat()[..],
+			 );
+			 let subspace_id: SpaceIdOf = generate_space_id::<T>(&sub_id_digest);
+
+			 let auth_id_digest = <T as frame_system::Config>::Hashing::hash(
+				 &[&subspace_id.encode()[..], &did.encode()[..]].concat()[..],
+			 );
+			 let authorization_id: AuthorizationIdOf = generate_authorization_id::<T>(&auth_id_digest);
+
+			 let origin =  <T as Config>::EnsureOrigin::generate_origin(caller, did);
```
vatsa287: Change `did` to `did.clone()`

vatsa287 (changes requested):

pallets/chain-space/src/benchmarking.rs
amarts: thanks. yes, indeed that was the issue.

vatsa287 (approved):
LGTM

PR #331, opened 2024-02-13
pallet/asset: Add Asset benchmark

vatsa287:
Add benchmark for pallet/asset.
Mock file is also added since it is a dependency on benchmark.

Signed-off-by: Shreevatsa N <vatsa@dhiway.com>


pallets/asset/src/tests.rs
```diff
@@ -0,0 +1,84 @@
+use super::*;
```
amarts: does this need `#[cfg(test)]` in lib.rs?

vatsa287: Yes it is required. Have added in lib.rs

amarts (approved):
LGTM

PR #334, opened 2024-02-20
GH-Workflow: Disable cargo build until issue is fixed

vatsa287:
Disable github cargo build workflow until the issue 'out of space' is fixed. Since clippy also does cargo build, disable it.

Signed-off-by: Shreeevatsa N <vatsa@dhiway.com>


amarts (approved):
we will need to work on better CI. But anyways, our test cases are getting improved, so I see that we are able to cover multiple pallet build and tests better. This should be fine to accept for now.

Issue #323, opened 2024-02-08
Identifier: Recheck InvalidIdentifier error code in `Identifier`


vatsa287:
## Description
During creation of a `Identifier` in cord-identifier package, recheck error codes being returned even when input data length is within defined range.

## Goals

- [ ] Correct error codes being returned.
- [ ] Recheck calculation of input data length.

## Expected Outcome

- [ ] Creation of valid identifier within defined range, with correct error codes otherwise.

## Acceptance Criteria
NA

## Implementation Details
NA

## Mockups / Wireframes
NA

---- 

## Product Name
CORD

## Organization Name
Dhiway

## Domain
Blockchain

## Tech Skills Needed
Rust

## Mentor(s)
@smohan-dw @amarts 

## Complexity
[Medium]

## Category
[Bug], [Test]

## Sub Category
[Backend], [Research]

Harsh1s:
Hello there, I investigated into this issue and found the culprit. Before I make a PR, just wanted to keep you in the loop and wanted your advice on the fix. Here's what I did and found:

Here's what I've found:

I added a debugging print statement at line 189 to print the length of the Base58 string converted from vector v: `println!("Length of Base58 string: {}", v.to_base58());`

Having run the test from my unmerged PR #336 with length 40 for v, The length of the Base58 string is 62, which is longer than the maximum allowed length for an identifier (`MAXIMUM_IDENTIFIER_LENGTH_U32`), which is 49.

The reason the Base58 string is longer is due to the nature of Base58 encoding. Base58 encoding is a type of binary-to-text encoding designed to be human-readable and avoid similar-looking characters. However, this encoding tends to increase the length of the encoded string compared to the original binary data.

The InvalidIdentifier error is thrown when the Base58 string is converted into a `BoundedVec<u8, ConstU32<MAXIMUM_IDENTIFIER_LENGTH_U32>>` at line 98. The BoundedVec type is a vector that has a compile-time fixed maximum length. If the length of the Vec<u8> (derived from the Base58 string) exceeds this maximum length, the try_into() call fails and returns an `InvalidIdentifier` error.

TLDR, We're having a couple checks for valid length, one with the input vector and one with the encoded Base58 string. 

Here are the two possible solutions I propose:
1. We get rid of the Base58 string max length check and rely altogether on the existing input vec length check.
2. We create a new max length limit for Base58 encoded string that is appropriate after the encoding.

[...]

PR #336, opened 2024-02-20
Feat: Adds test for `InvalidIdentifier`

### LINKED ISSUE #279: [C4GT]  Chain-Space: Add tests for `InvalidIdentifier`
## Description

New testcase can be tested under `cargo test -p pallet-chain-space` after adding it in the code.

## Goals

- [ ] Add tests for `InvalidIdentifier` for pallet/chain-space

## Expected Outcome
- [ ] Test should assert for `InvalidIdentifier` being returned properly in all the possible calls.
       
## Acceptance Criteria
NA

## Implementation Details
Ex: `creating_a_duplicate_space_should_fail` which validates error code `SpaceAlreadyAnchored`

## Mockups / Wireframes
NA

---- 

## Product Name
CORD

## Organization Name
Dhiway

## Domain
Blockchain

## Tech Skills Needed
Rust

## Mentor(s)
@amarts 

## Complexity
[Medium]

## Category
[Test]

## Sub Category
[Beginner friendly]

Harsh1s:
This PR implements test-case for error-code `InvalidIdentifier` for chain-space pallet.
When invalid base58 encoded data is used, it gives this error-code.

Fixes: #279 

I'm just starting with the project code base hence am not so confident. I hope I did the correct thing, if not, I'll try my best to learn and do it again. Thanks for your patience!

### DIFF
--- a/pallets/asset/src/lib.rs
+++ b/pallets/asset/src/lib.rs
@@ -27,6 +27,9 @@ pub mod mock;
 #[cfg(feature = "runtime-benchmarks")]
 pub mod benchmarking;
 
+#[cfg(test)]
+mod tests;
+
 pub use crate::{pallet::*, types::*};
 use frame_support::{ensure, traits::Get};
 use identifier::{
--- a/pallets/asset/src/tests.rs
+++ b/pallets/asset/src/tests.rs
@@ -1,0 +1,157 @@
+use super::*;
+use crate::mock::*;
+use codec::Encode;
+use cord_utilities::mock::{mock_origin::DoubleOrigin, SubjectId};
+use frame_support::{assert_ok, BoundedVec};
+use frame_system::RawOrigin;
+use pallet_chain_space::{SpaceCodeOf, SpaceIdOf};
+use sp_runtime::{traits::Hash, AccountId32};
+use sp_std::prelude::*;
+
+/// Generates a space ID from a digest.
+pub fn generate_space_id<T: Config>(digest: &SpaceCodeOf<T>) -> SpaceIdOf {
+	Ss58Identifier::create_identifier(&(digest).encode()[..], IdentifierType::Space).unwrap()
+}
+
+/// Generates an authorization ID from a digest.
+pub fn generate_authorization_id<T: Config>(digest: &SpaceCodeOf<T>) -> AuthorizationIdOf {
+	Ss58Identifier::create_identifier(&(digest).encode()[..], IdentifierType::Authorization)
+		.unwrap()
+}
+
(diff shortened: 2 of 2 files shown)


pallets/chain-space/src/tests.rs
```diff
@@ -4,6 +4,7 @@ use codec::Encode;
 use cord_utilities::mock::{mock_origin::DoubleOrigin, SubjectId};
 use frame_support::{assert_err, assert_ok, error::BadOrigin};
 use frame_system::RawOrigin;
+use identifier::IdentifierError;
```
vatsa287: This will be part of identifier tests.
Move it to primitives/identifier module.

pallets/chain-space/src/tests.rs
```diff
@@ -1082,3 +1083,22 @@ fn restoring_approval_of_a_space_by_non_root_should_fail() {
 		);
 	});
 }
+
+#[test]
+fn test_invalid_identifier_error() {
+	let invalid_base58_data = "I am not a Base58 encoded string".as_bytes().to_vec();
+	let valid_base58_data = [2u8; 30].to_vec();
```
vatsa287: When this is changed to 40, which is still a valid length of identifier, still the error code "Invalid Identifier" is thrown. So this requires more investigation. 
Try with `let valid_base58_data = [2u8; 40].to_vec();`
Check issue #323

Harsh1s: I investigated a bit and found something weird. I used the length 30 as it was used in the test `creating_a_invalid_identifier_length_should_fail` too. But here's the thing, if I change it to 40 even in the `creating_a_invalid_identifier_length_should_fail` test, it fails. I'm quite new to the code base so I'm not sure if that's supposed to happen. Would like to hear your thoughts on this.

vatsa287: Yes that is what needs to be solved with #323. No errors should be propagated if identifier is within VALID length.

Harsh1s: I'll try solving both these issues then. Could you please help me understand what exactly is causing the `invalidIdentifierLength` error-code even when it is well under the valid range in case of 40?

Harsh1s: Hey @vatsa287. I found the culprit for #323 and will make a PR to solve that one after hearing your thoughts on the same. I have listed my findings as a comment on the issue. I'll make a PR for #279 after that.

amarts: Cool. Good effort and check there. Please send the PR.

PR #342, opened 2024-02-27
Chain-Space: Add tests for SpaceDelegatesLimitExceeded

### LINKED ISSUE #281: [C4GT] Chain-Space: Add tests for `SpaceDelegatesLimitExceeded`
## Description

New testcase can be tested under `cargo test -p pallet-chain-space` after adding it in the code.

## Goals

- [ ] Add tests for `SpaceDelegatesLimitExceeded` for pallet/chain-space

## Expected Outcome
- [ ] Test should assert for `SpaceDelegatesLimitExceeded` being returned properly in all the possible calls.

## Acceptance Criteria
NA

## Implementation Details
Ex: `creating_a_duplicate_space_should_fail` which validates error code `SpaceAlreadyAnchored`

## Mockups / Wireframes
NA

---- 

## Product Name
CORD

## Organization Name
Dhiway

## Domain
Blockchain

## Tech Skills Needed
Rust

## Mentor(s)
@amarts 

## Complexity
[Low]

## Category
[Test]

## Sub Category
[Beginner friendly]

Harsh1s:
Fixes #281 

Details:
Two functions return this error code, namely `Space::create` and `Space::space_delegate_addition`. Used these two for this test.

### DIFF
--- a/pallets/chain-space/src/tests.rs
+++ b/pallets/chain-space/src/tests.rs
@@ -1105,16 +1105,13 @@ fn add_delegate_should_fail_if_space_delegates_limit_exceeded() {
 
 		// Add the maximum number of delegates to the space
 		let mut delegate_count: u8 = 0;
-		loop {
-			match Space::space_delegate_addition(
-				space_id.clone(),
-				SubjectId(AccountId32::new([delegate_count; 32])),
-				creator.clone(),
-				Permissions::all(),
-			) {
-				Ok(_) => delegate_count += 1,
-				Err(_) => break,
-			}
+		while let Ok(_) = Space::space_delegate_addition(
+			space_id.clone(),
+			SubjectId(AccountId32::new([delegate_count; 32])),
+			creator.clone(),
+			Permissions::all(),
+		) {
+			delegate_count += 1;
 		}
 
 		// Attempt to add one more delegate, which should exceed the limit and result in the
--- a/pallets/chain-space/src/tests.rs
+++ b/pallets/chain-space/src/tests.rs
@@ -1104,14 +1104,13 @@ fn add_delegate_should_fail_if_space_delegates_limit_exceeded() {
 		));
 
 		// Add the maximum number of delegates to the space
-		let mut delegate_count: u8 = 0;
-		while let Ok(_) = Space::space_delegate_addition(
-			space_id.clone(),
-			SubjectId(AccountId32::new([delegate_count; 32])),
-			creator.clone(),
-			Permissions::all(),
-		) {
-			delegate_count += 1;
+		for delegate_count in 0..4 {
+			assert_ok!(Space::space_delegate_addition(
+				space_id.clone(),
+				SubjectId(AccountId32::new([delegate_count; 32])),
+				creator.clone(),
+				Permissions::all(),
+			));
 		}
 
 		// Attempt to add one more delegate, which should exceed the limit and result in the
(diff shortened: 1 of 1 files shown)


vatsa287:
@Harsh1s perform rust format with,
`rustfmt <file>` & cargo test with `cargo test -p pallet-chain-space`.

See Readme on pre-PR checks.

Harsh1s:
> @Harsh1s perform rust format with, `rustfmt <file>` & cargo test with `cargo test -p pallet-chain-space`.
> 
> See Readme on pre-PR checks.

Thanks @vatsa287 for the help, I've fixed it now.

Harsh1s:
> > @Harsh1s perform rust format with, `rustfmt <file>` & cargo test with `cargo test -p pallet-chain-space`.
> > 
> > See Readme on pre-PR checks.
> 
> Thanks @vatsa287 for the help, I've fixed it now.

Seems like using the loop failed the test with clippy. I'll replace that too.

pallets/chain-space/src/tests.rs
```diff
@@ -1082,3 +1082,48 @@ fn restoring_approval_of_a_space_by_non_root_should_fail() {
...
+
+	let id_digest = <Test as frame_system::Config>::Hashing::hash(
+		&[&space_digest.encode()[..], &creator.encode()[..]].concat()[..],
+	);
+
+	let space_id: SpaceIdOf = generate_space_id::<Test>(&id_digest);
+
+	new_test_ext().execute_with(|| {
+		// Create the space
+		assert_ok!(Space::create(
```
vatsa287, line 1101: Even though space::create() has check for `spaceDelegatesLimitExceeded`, It is not possible hit that case since creation of every chain space is unique, based on creator & space_id.

pallets/chain-space/src/tests.rs
```diff
@@ -1082,3 +1082,48 @@ fn restoring_approval_of_a_space_by_non_root_should_fail() {
...
+			DoubleOrigin(author.clone(), creator.clone()).into(),
+			space_digest,
+		));
+
+		// Add the maximum number of delegates to the space
+		let mut delegate_count: u8 = 0;
+		while let Ok(_) = Space::space_delegate_addition(
```
vatsa287: Since `MaxSpaceDelegates` is set at 10_000 in runtime, this loop might run that many times if in runtime.
It is ok if there is no other way to change just for tests as in mock.

vatsa287 (dismissed):
LGTM

pallets/chain-space/src/tests.rs
amarts: I guess it is OK to reduce the max delegates to 4-5 in mock. All tests can finish within that many delegates for now.

vatsa287: https://github.com/dhiway/cord/blob/develop/pallets/chain-space/src/mock.rs#L85
It is set to 5.

amarts (commented):
LGTM.

Minor comment - While the tests would execute fine here, should we rather test if delegate_count is 5 instead of looping till we get error? Ie, if there is a bug we would continue the loop till u64 times? 

A fixed loop for max delegate count would be a better model IMO.

pallets/chain-space/src/tests.rs
Harsh1s: Thanks for the info, makes great sense tbh. Thanks!

pallets/chain-space/src/tests.rs
Harsh1s: I could add 4 delegates after which I got an error (creator was the 1st delegate ig hence there was space for 4 more), so I have modified the code to for loop 4 times to fill the space.

Harsh1s:
@amarts thanks for your help, I understand the project better now! I have made changes just as you suggested and fully agree with you on this.

vatsa287 (approved):

PR #349, opened 2024-03-01
sdk v1.8.0 update

smohan-dw:


runtime/src/lib.rs
```diff
@@ -109,7 +109,7 @@ include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));
 /// Max size for serialized extrinsic params for this testing runtime.
 /// This is a quite arbitrary but empirically battle tested value.
 #[cfg(test)]
-pub const CALL_PARAMS_MAX_SIZE: usize = 240;
+pub const CALL_PARAMS_MAX_SIZE: usize = 244;
```
amarts, line 112: nitpick: only change i see in the PR, which doesn't have any corresponding requirement mentioned.

test-utils/runtime/src/genesismap.rs
```diff
@@ -61,6 +61,7 @@ impl Default for GenesisStorageBuilder {
 				Sr25519Keyring::Charlie.into(),
 			],
 			(0..16_usize)
+				.into_iter()
```
amarts, line 64: is this anticipated? How was it working earlier?

amarts (approved):
2 minor comments, can be looked into as a 'comment' to PR as I don't see any major codechange required here.

amarts:
Will merge this, so it helps to base everything to latest base. Lets take any other PR on top of this.

PR #355, opened 2024-03-05
did: remove the did counter check for did_submit_call()

### LINKED ISSUE #186: `api.tx.utility.batchAll()` error occurrence
**Issue:**
Encountering an error "The DID operation nonce does not match the current DID nonce + 1" when using `api.tx.utility.batchAll()`.

**Requirement**:
For bulk uploads or revocations of records in the studio.

amarts:
Fixes: https://github.com/dhiway/cord.js/issues/186

### DIFF
--- a/pallets/did/src/lib.rs
+++ b/pallets/did/src/lib.rs
@@ -1181,7 +1181,10 @@ pub mod pallet {
 			let mut did_details = Did::<T>::get(&operation.did)
 				.ok_or(StorageError::NotFound(errors::NotFoundKind::Did))?;
 
-			Self::validate_counter_value(operation.tx_counter, &did_details)?;
+			// TODO: With agency model this nonce check would cause a problem when we get
+			// to utility batch etc. Re-enable this once we have more tests
+			// Self::validate_counter_value(operation.tx_counter, &did_details)?;
+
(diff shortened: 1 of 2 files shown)


amarts:
Discuss on pros and cons of this change before merging.

amarts:
Cons: 
* Without nonce check, the order of calls is not strictly checked.
* Without nonce check, the pallets will have to handle the duplicate calls if any (made from sdk/rpc).

Pros:
* With identifier based approach we have, the duplicate call issue is not a major one.
* Helps to scale and use same 'agency' account to sign from multiple machines for handling the calls / operations in parallel, so we can scale to more number of operations per block from same author.
* Considering we haven't removed the counter increments, this would help anyways.

amarts:
this strict check on 'did.tx_count + 1' may not be required, as it may be completely stopping the 'utility.batch()' usecases.

Keep this check commented for a release and re-enable once its well tested.

vatsa287 (approved):
I'll open a issue to track this requirement.

amarts:
Comments from @smohan-dw 

> The counter check is linked to DID signature and is a measure the check the signatures are not reused. This would prevent a replay attack on the chain.

Will revert this, and as updated in the cord.js issue, will work on available framework in DID methods to get the batch working.

PR #373, opened 2024-03-12
Identity: Add tests for RegistrarAlreadyExists Issue: 289 modified

Sriramkashyapks:
ive added the file which is necessary according to my previous suggestion


vatsa287:
Please run `rustfmt pallets/identity/src/tests.rs` for rust format to pass

PR #389, opened 2024-03-15
pallet/asset: New asset extrinsic methods for VC

vatsa287:
This PR,
- Creates 4 new extrinsic methods `vc_create`, `vc_issue`, `vc_transfer` & `vc_status_change`  for VC create & issue.
- Pallet storage changes.
- Asset type changes.
- Unit tests to demonstrate the same.

Signed-off-by: Shreevatsa N <vatsa@dhiway.com>

### DIFF
--- a/pallets/asset/src/types.rs
+++ b/pallets/asset/src/types.rs
@@ -41,15 +41,6 @@ pub struct AssetInputEntry<AssetDescription, AssetTypeOf, AssetTag, AssetMeta> {
 	/// open structure - 1024 bytes max
 	pub asset_meta: AssetMeta,
 }
-
-#[derive(
-	Encode, Decode, Clone, RuntimeDebug, PartialEq, Eq, PartialOrd, Ord, TypeInfo, MaxEncodedLen,
-)]
-pub struct VCAssetInputEntry {
-	/// asset quantity
-	pub asset_qty: u64,
-}
-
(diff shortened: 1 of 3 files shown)


amarts (commented):
@vatsa287 considering `<VCIssuance>` is a separate one, we need a 'vc_transfer' as a call too. (where you check if the VCInstance is valid, and pick transfer from there). Other things seems fine so far.

amarts:
another thing to work on is, with this code built locally, run `cord --dev` and get asset-tx script to run with newer calls from cord.js

vatsa287:
> @vatsa287 considering `<VCIssuance>` is a separate one, we need a 'vc_transfer' as a call too. (where you check if the VCInstance is valid, and pick transfer from there). Other things seems fine so far.

Added `vc_transfer` & `vc_status_change`. Since type of Instance we store with VCInstance with storage will change.

pallets/asset/src/types.rs
```diff
@@ -38,6 +42,14 @@ pub struct AssetInputEntry<AssetDescription, AssetTypeOf, AssetTag, AssetMeta> {
 	pub asset_meta: AssetMeta,
 }
 
+#[derive(
```
vatsa287: @amarts I will discard `VCAssetInputEntry` & `VCAssetInputEntryOf` in lib.rs as asset_qty will be taken as an input parameter directly.

amarts (commented):
LGTM

amarts (approved):
LGTM

PR #393, opened 2024-03-17
identifier: Network ID in create identifier

amarts:
This is not yet ready, and there is a compilation error.

The main thing to note with this effort is:

* Ability to pass a ss58IdentifierFormat from bootstrap file -> chain_spec file.

* ChainSpec file adds the fields to genesis_block/genesis_build of identifier and passes this information to pallet.

* With this, a network will always have identifier with same prefix and hence similar identifiers

### DIFF
--- a/node/cli/src/chain_spec.rs
+++ b/node/cli/src/chain_spec.rs
@@ -132,6 +132,7 @@ fn cord_development_config_genesis() -> serde_json::Value {
 			get_account_id_from_seed::<sr25519::Public>("Alice"),
 		)],
 		get_account_id_from_seed::<sr25519::Public>("Alice"),
+		29,
 	)
 }
 
@@ -161,6 +162,7 @@ fn cord_local_config_genesis() -> serde_json::Value {
 			),
 		],
 		get_account_id_from_seed::<sr25519::Public>("Alice"),
+		29,
 	)
 }
 
@@ -218,8 +220,12 @@ fn cord_local_genesis(
 	initial_authorities: Vec<(AccountId, BabeId, GrandpaId, ImOnlineId, AuthorityDiscoveryId)>,
 	initial_well_known_nodes: Vec<(NodeId, AccountId)>,
 	root_key: AccountId,
+	ss58_identifier_format: u16,
 ) -> serde_json::Value {
 	serde_json::json!( {
+		"identifier": {
+			"ss58IdentifierFormat": ss58_identifier_format,
+		},
 		"nodeAuthorization":  {
 			"nodes": initial_well_known_nodes.iter().map(|x| (x.0.clone(), x.1.clone())).collect::<Vec<_>>(),
 		},
--- a/node/cli/src/chain_spec.rs
+++ b/node/cli/src/chain_spec.rs
@@ -20,7 +20,7 @@
 
 pub mod bootstrap;
 
-pub use cord_primitives::{AccountId, Balance, NodeId, Signature};
+pub use cord_primitives::{AccountId, Balance, NodeId, Signature, DEFAULT_SS58_IDENTIFIER_PREFIX};
 pub use cord_runtime::RuntimeGenesisConfig;
 use cord_runtime::{Block, SessionKeys};
 use pallet_im_online::sr25519::AuthorityId as ImOnlineId;
@@ -132,7 +132,7 @@ fn cord_development_config_genesis() -> serde_json::Value {
 			get_account_id_from_seed::<sr25519::Public>("Alice"),
 		)],
 		get_account_id_from_seed::<sr25519::Public>("Alice"),
-		29,
+		DEFAULT_SS58_IDENTIFIER_PREFIX,
 	)
 }
 
@@ -162,7 +162,7 @@ fn cord_local_config_genesis() -> serde_json::Value {
 			),
 		],
 		get_account_id_from_seed::<sr25519::Public>("Alice"),
-		29,
+		DEFAULT_SS58_IDENTIFIER_PREFIX,
 	)
 }
 
--- a/node/cli/src/chain_spec/bootstrap.rs
+++ b/node/cli/src/chain_spec/bootstrap.rs
@@ -39,6 +39,7 @@ use crate::chain_spec::{get_properties, Extensions, CORD_TELEMETRY_URL, DEFAULT_
 pub struct ChainParams {
 	pub chain_name: String,
 	pub chain_type: ChainType,
+	pub ss58_identifier_format: u16,
 	pub authorities: Vec<Vec<String>>,
 	pub well_known_nodes: Vec<Vec<String>>,
 	pub network_members: Vec<String>,
@@ -54,6 +55,9 @@ impl ChainParams {
 
 	pub fn chain_name(&self) -> &str {
 		&self.chain_name
+	}
+	pub fn ss58_identifier_format(&self) -> u16 {
+		self.ss58_identifier_format
 	}
 }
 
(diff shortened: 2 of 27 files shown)


node/cli/src/command/chain_setup.rs
```diff
@@ -78,6 +79,11 @@ impl BootstrapChainCmd {
 				other
 			)),
 		};
+		let network_id = if config.n.len() <= 64 {
```
vatsa287: Maybe for now just set `let network_id = config.network_id.clone()`. We can add a check later.

scripts/bootstrap.sh
```diff
@@ -10,6 +10,8 @@ function usage() {
   echo "  -a, --num-authorities Number of authorities (validators)"
   echo "  -s, --SECRET          Secret key or seed for generating new accounts"
   echo "  -o, --output-directory Directory where output files will be saved (default is current directory)"
+  echo "  -i, --network         Network ID Prefix for ss58 identifiers"
```
vatsa287: Update above usage statement too, as in `-c <chain_name>` ..

scripts/bootstrap.sh
```diff
@@ -34,6 +38,8 @@ while getopts "m:n:a:s:o:" flag; do
   a) NUM_AUTHORITIES=${OPTARG} ;;
   s) SECRET=${OPTARG} ;;
   o) OUTPUT_DIR=${OPTARG} ;;
+  i) NETWORK_ID=${OPTARG} ;;
```
vatsa287: Also add `:i:c` as a flag above in the while statement.

node/cli/src/chain_spec/bootstrap.rs
```diff
@@ -116,11 +121,13 @@ fn cord_custom_config_genesis(config: ChainParams) -> serde_json::Value {
 		initial_council_members,
 		initial_tech_committee_members,
 		initial_sudo_key,
+		prefix,
```
vatsa287: Have to set `prefix` as `network_id` which in-turn gets passed to create `serde_json`.
Or use a variable for consistency `prefix = network_id.clone()`

vatsa287 (changes requested):
I see use of different types for prefix(u16) & network(u32). I did not get since `prefix` is derived from `network_id` itself, or they both have different use-case, like `chain_name: "cord", network_id: 01, prefix: 29`?

node/cli/src/chain_spec.rs
```diff
@@ -132,6 +132,7 @@ fn cord_development_config_genesis() -> serde_json::Value {
 			get_account_id_from_seed::<sr25519::Public>("Alice"),
 		)],
 		get_account_id_from_seed::<sr25519::Public>("Alice"),
+		29,
```
smohan-dw: Chain-spec will not allow the creation of a new variable at the top level. The only place to insert a new variable is via pallet config.

amarts: @smohan-dw would like to understand this a bit more. I was able to test getting the prefix loaded up in genesis_build call in pallet through this logic.

`2024-03-18 08:08:37.151  INFO main cord_identifier::pallet: Setting new prefix key 29`

(this log came with latest PR, which was done just to highlight passing of values from chain_spec perspective).

node/cli/src/chain_spec.rs
```diff
@@ -218,8 +220,12 @@ fn cord_local_genesis(
 	initial_authorities: Vec<(AccountId, BabeId, GrandpaId, ImOnlineId, AuthorityDiscoveryId)>,
 	initial_well_known_nodes: Vec<(NodeId, AccountId)>,
 	root_key: AccountId,
+	prefix: u16,
```
smohan-dw: ss58IdentifierFormat

node/cli/src/chain_spec.rs
```diff
@@ -161,6 +162,7 @@ fn cord_local_config_genesis() -> serde_json::Value {
 			),
 		],
 		get_account_id_from_seed::<sr25519::Public>("Alice"),
+		29,
```
smohan-dw: good to have this mapped to a variable

node/cli/src/chain_spec/bootstrap.rs
```diff
@@ -39,6 +39,7 @@ use crate::chain_spec::{get_properties, Extensions, CORD_TELEMETRY_URL, DEFAULT_
 pub struct ChainParams {
 	pub chain_name: String,
 	pub chain_type: ChainType,
+	pub network_id: u16,
```
smohan-dw: ss58IdentifierFormat

node/cli/src/chain_spec/bootstrap.rs
```diff
@@ -71,6 +75,7 @@ fn session_keys(
 
 /// Custom config.
 fn cord_custom_config_genesis(config: ChainParams) -> serde_json::Value {
+	let prefix = config.network_id();
```
smohan-dw: ss58IdentifierFormat

node/cli/src/chain_spec/bootstrap.rs
```diff
@@ -116,6 +121,7 @@ fn cord_custom_config_genesis(config: ChainParams) -> serde_json::Value {
 		initial_council_members,
 		initial_tech_committee_members,
 		initial_sudo_key,
+		prefix,
```
smohan-dw: ss58IdentifierFormat

node/cli/src/chain_spec/bootstrap.rs
```diff
@@ -147,8 +153,12 @@ fn cord_custom_genesis(
 	council_members: Vec<AccountId>,
 	tech_committee_members: Vec<AccountId>,
 	sudo_key: AccountId,
+	prefix: u16,
```
smohan-dw: ss58IdentifierFormat

node/cli/src/command/chain_setup.rs
```diff
@@ -79,8 +80,11 @@ impl BootstrapChainCmd {
 			)),
 		};
 
-		let chain_type = chain_type?;
```
smohan-dw, line 82: why are we removing this?

node/cli/src/command/chain_setup.rs
```diff
@@ -79,8 +80,11 @@ impl BootstrapChainCmd {
...
-
+		let network_id: u16 = if config.network_id != 0 { 
+			config.network_id
+		} else {
+			10029
```
smohan-dw: set a primitive variable for default

primitives/identifier/src/curi.rs
```diff
@@ -128,22 +88,17 @@ pub trait IdentifierCreator {
...
-pub trait CordIdentifierType {
-	fn get_type(&self) -> Result<IdentifierType, IdentifierError>;
-}
-
-impl CordIdentifierType for Ss58Identifier {
-	fn get_type(&self) -> Result<IdentifierType, IdentifierError> {
-		let identifier_type_u16 = self.get_identifier_type()?;
-
-		IdentifierType::from_u16(identifier_type_u16).ok_or(IdentifierError::InvalidIdentifier)
+		Ss58Identifier::from_encoded(data, 10029)
```
smohan-dw: prefix value should be read from the storage

scripts/bootstrap.sh
amarts: Done

node/cli/src/chain_spec.rs
amarts: Done

PR #394, opened 2024-03-17
pallet identity updates with flexible fields and names

smohan-dw:


amarts (dismissed):
Code looks fine. Some minor info about added username would be good to have in commit description.

amarts (dismissed):

amarts (approved):

Issue #377, opened 2024-03-12
[C4GT] Asset: Add tests for `InstanceNotActive`


vatsa287:
## Description
Subtask under : https://github.com/cord-network/community/issues/7

New testcase can be tested under `cargo test -p pallet-asset` after adding it in the code.

## Goals

- [ ] Add tests for `InstanceNotActive` for pallet/asset

## Expected Outcome
- [ ] Test should assert for `InstanceNotActive` being returned properly in all the possible calls.
- [ ] If there are more than one function returning this error code, all of them should be validated either as one test case or as multiple test cases.
       
## Acceptance Criteria
NA

## Implementation Details
Look at other test cases, and add a test case for the same.
Ex: check method `asset_create_should_succeed` which validates creation of asset.

## Mockups / Wireframes
NA

---- 

## Product Name
CORD

## Organization Name
Dhiway

## Domain
Blockchain

## Tech Skills Needed
Rust

## Mentor(s)
@amarts @vatsa287 

## Complexity
[Medium]

## Category
[Test]

## Sub Category
[Beginner friendly]

Sriramkashyapks:
please assign this issue to me

summerpun2803:
@vatsa287  can i work on this issue?

VedantKhairnar:
Hello @vatsa287 
If the PR is merged for this issue, can we close this one?
Thanks.

vatsa287:
Closing as completed from #483.

PR #395, opened 2024-03-18
added tests for InstanceNotActive #377 in pallet-asset

Sriramkashyapks:


pallets/asset/src/tests.rs
```diff
@@ -155,3 +155,83 @@ fn asset_issue_should_succeed() {
...
+
+	new_test_ext().execute_with(|| {
+		assert_ok!(Space::create(
+			DoubleOrigin(author.clone(), creator.clone()).into(),
+			space_digest,
+		));
+
+		assert_ok!(Space::approve(RawOrigin::Root.into(), space_id, capacity));
+
+		assert_ok!(Asset::create(
```
vatsa287, line 219: Creation of an asset will be with status: `active`.

pallets/asset/src/tests.rs
```diff
@@ -155,3 +155,83 @@ fn asset_issue_should_succeed() {
...
+
+		// Now, try to issue the asset to the inactive instance
+		assert_noop!(
+			Asset::issue(
+				DoubleOrigin(author.clone(), creator.clone()).into(),
+				issue_entry.clone(),
+				issue_entry_digest,
+				authorization_id
+			),
+			Error::<Test>::InstanceNotActive
```
vatsa287: This would give `Ok()`, since this is a valid call to issue an asset.

vatsa287 (changes requested):
@Sriramkashyapks I do not see `InstanceNotActive` being tested here. Please check for that error code.

Issue #375, opened 2024-03-12
[C4GT] Asset: Add tests for `AssetIdNotFound`


vatsa287:
## Description
Subtask under : https://github.com/cord-network/community/issues/7

New testcase can be tested under `cargo test -p pallet-asset` after adding it in the code.

## Goals

- [ ] Add tests for `AssetIdNotFound` for pallet/asset

## Expected Outcome
- [ ] Test should assert for `AssetIdNotFound` being returned properly in all the possible calls.
- [ ] If there are more than one function returning this error code, all of them should be validated either as one test case or as multiple test cases.
       
## Acceptance Criteria
NA

## Implementation Details
Look at other test cases, and add a test case for the same.
Ex: check method `asset_create_should_succeed` which validates creation of asset.

## Mockups / Wireframes
NA

---- 

## Product Name
CORD

## Organization Name
Dhiway

## Domain
Blockchain

## Tech Skills Needed
Rust

## Mentor(s)
@amarts @vatsa287 

## Complexity
[Low]

## Category
[Test]

## Sub Category
[Beginner friendly]

aishalxgupta:
Hi @vatsa287 
I'm confident in my understanding of the problem and possess the necessary skills. I'd be grateful for the opportunity to be assigned to this project

vatsa287:
Hi @aishalxgupta Welcome. You can work on this issue. You can refer other PRs of C4GT. Also can connect through C4GTs discord channel.

vatsa287:
Closing as completed.

PR #397, opened 2024-03-20
added test for AssetIdNotFound #375 in pallet-asset

aishalxgupta:
- This test case creates a space, authorization ID, and an asset definition.

- It then creates an issue_entry with a non-existent asset_id using a different digest.

- The test calls Asset::issue with this invalid issue_entry and asserts that the result is an error containing AssetIdNotFound.


vatsa287:
@aishalxgupta Any update on this PR. Pleas rebase and update.
Changes to be made:
- Check for `AssetIdNotFound` in all calls where this error code is used as in `issue, transfer, status_change, vc_issue, vc_transfer & vc_status_change`
- Run rustfmt on all changed files for the Rust Format test to pass.

PR #422, opened 2024-04-08
Added Test for RatingIdentifierAlreadyExist

zeel991:
- In this Test , the when the `registered rating` is revoked , it should give `RatingIdentifierAlreadyExist` Error if the Identifer is already in use .
- Since the `EntryTypeOf` when set to Credit , gave a different error , I had to manually set `EntrytypeOf` to `Debit` 
@vatsa287 Please review the PR and give your Feedback .
Thank you .

### DIFF
--- a/pallets/network-score/src/tests.rs
+++ b/pallets/network-score/src/tests.rs
@@ -308,3 +308,79 @@ fn test_register_rating_id_already_exists() {
 		);
 	});
 }
+
+#[test]
+fn test_revoke_rating_id_already_exists() {
+	let creator = DID_00.clone();
+	let author = ACCOUNT_00.clone();
+	let message_id = BoundedVec::try_from([72u8; 10].to_vec()).unwrap();
+	let entity_uid = BoundedVec::try_from([73u8; 10].to_vec()).unwrap();
+	let provider_uid = BoundedVec::try_from([74u8; 10].to_vec()).unwrap();
+	let entry = RatingInputEntryOf::<Test> {
+		entity_uid,
+		provider_uid,
+		total_encoded_rating: 250u64,
+		count_of_txn: 7u64,
+		entity_type: EntityTypeOf::Logistic,
+		rating_type: RatingTypeOf::Overall,
+		provider_did: creator.clone(),
+	};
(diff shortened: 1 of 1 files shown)


zeel991:
Hey @vatsa287  should I use `Match` in this way 
```
 match RatingEntries::<Test>::get(&identifier) {
            Some(rating_entry) => {
                // Unwrap the Option to access the inner RatingEntry struct
                let mut rating_entry = rating_entry;
                .
                .
                .
                assert_err!(
                    Score::revise_rating(
                        DoubleOrigin(author.clone(), creator.clone()).into(),
                       .
                       .
                       .
                    ),
                    Error::<Test>::RatingIdentifierAlreadyAdded
                );
            },
            None => {
                Error::<Test>::RatingIdentifierNotFound
            }
     }
```
Please review the PR .

zeel991:
Hey @vatsa287 , Can you please review the PR

pallets/network-score/src/tests.rs
```diff
@@ -312,3 +312,90 @@ fn test_register_rating_id_already_exists() {
...
+		));
+
+		// Remove message_id and provider_did from entries
+		<MessageIdentifiers<Test>>::remove(message_id.clone(), creator.clone());
+		// Entry type of should not be anything other than Debit
+		rating_entry.entry_type = EntryTypeOf::Debit;
+		RatingEntries::<Test>::insert(&identifier, &rating_entry);
+
+		assert_err!(
+			Score::revise_rating(
```
vatsa287: Revise rating expects entry type to be of `Debit` in the storage. To test this can be done by 
- Calling `revoke_rating` which does exactly this ( Ideal ).
- Setting entry-type to be of `DEBIT` in storage ( Not Recommended in this case since we can do through a method from above (revoke_rating()).

You can try again, I'll post entire solution tomorrow.

zeel991: Ohk @vatsa287 , Which method from 1 ?

vatsa287: @zeel991 execute `revoke_rating()`

PR #432, opened 2024-04-19
disable block authoring backoff on production networks

smohan-dw:
Also, starting with the runtime renaming, Braid will be the default name for the tokenless enterprise network runtimes. There will be no change to the binary name.


amarts (commented):
LGTM

One question: What would be the 'spec' name when the typebundle is shared with cord.js? Because the typebundle initialization in CORD.js side happens with spec name == `cord` right now. If that changes, we need to have corresponding change implemented.

smohan-dw:
> LGTM
> 
> One question: What would be the 'spec' name when the typebundle is shared with cord.js? Because the typebundle initialization in CORD.js side happens with spec name == `cord` right now. If that changes, we need to have corresponding change implemented.

spec is now mapped with protocol-id which is not changing. A better fix at the SDK level is to separate the augment-api package with multi-chain support.

smohan-dw:
The changes made to the chain ID names have been reverted. The chain IDs will continue to be referred to as `dev`, `local`, and `live`.

amarts (approved):

PR #434, opened 2024-04-25
asset [Minor]: Add checks for InvalidAssetStatus

vatsa287:
There already exists a trait for `is_valid_status_type`. Add a new error enum & add checks for incoming status using the trait.


smohan-dw (approved):
LGTM

Issue #312, opened 2024-02-06
 [C4GT] Schema: Add tests for `MaxEncodedSchemaLimitExceeded`


vatsa287:
## Description
Subtask under : https://github.com/cord-network/community/issues/7

New testcase can be tested under `cargo test -p pallet-schema` after adding it in the code.

## Goals

- [ ] Add tests for `MaxEncodedSchemaLimitExceeded` for pallet/schema

## Expected Outcome
- [ ] Test should assert for `MaxEncodedSchemaLimitExceeded` being returned properly in all the possible calls.
- [ ] If there are more than one function returning this error code, all of them should be validated either as one test case or as multiple test cases.
       
## Acceptance Criteria
NA

## Implementation Details
Look at other test cases, and add a test case for the same.
Ex: check method `check_duplicate_schema_creation` which validates error code `SchemaAlreadyAnchored`

## Mockups / Wireframes
NA

---- 

## Product Name
CORD

## Organization Name
Dhiway

## Domain
Blockchain

## Tech Skills Needed
Rust

## Mentor(s)
@amarts 

## Complexity
[Medium]

## Category
[Test]

## Sub Category
[Beginner friendly]

shashank-iitbhu:
I don't think it is possible to create a schema exceeding `MaxEncodedSchemaLength` and pass it to the create function.
The `InputSchemaOf<T>` type is defined as a `BoundedVec<u8, <T as Config>::MaxEncodedSchemaLength>`, which means it is inherently limited to a maximum size specified by `MaxEncodedSchemaLength`.

So I don't think there is a need to write an explicit test for this. What do you say? or am i missing something?
@vatsa287

shriyanyk:
Hi i would like to work on this issue.

Pankajjsuthar:
Hi, I was currently working on this issue and find out that "schema size should be max of MaxEncodedSchemaLength and that we are getting as an input and we have to make tests which make sure that schema doesn't exceed the limit" am I right here or am i missing some details here. Can you please tell me more so that I can work on these issues, I think there are 10 in same way, so if I do one in good manner others will be easier for me.

zeel991:
Hey @Pankajjsuthar , as you can see there is already a PR linked to the issue, so you can work on different issue

VedantKhairnar:
Hello @Pankajjsuthar Are you open to working on this issue?
We will be happy to assign it to you now.
Thanks.

Vaibhavsahu2810:
[...]

PR #435, opened 2024-04-25
Improved Tests for Schema Size Limit (Issue #312)(Adding Test max encoded schema limit exceeded to pallet/schema)

HarishKarthickS:
Hi everyone,

This pull request addresses issue #312 and aims to improve the testing of the maximum schema size limit. These improvements ensure that the system correctly handles scenarios where the schema size might be exceeded.

My Changes:

I've added more comprehensive tests to the existing suite:
One test specifically checks the Schema::create function to verify it throws the MaxEncodedSchemaLimitExceeded error when data is too large.
Another test case focuses on an edge case, making sure the limit check works correctly for data sizes very close to the maximum allowed size.
I've used a mock storage system (MockStorage) to isolate the tests from external dependencies, making them more reliable.
I've included comments throughout the code to explain the purpose of each section (arrange, act, assert) for better understanding.
Benefits:

These additional tests will help us catch potential issues related to exceeding the schema size limit before they occur in production, leading to a more robust system.
Developers can be more confident in how the system handles large schemas.
Clear test cases will make it easier to maintain the code in the future.
Next Steps:

I'd appreciate it if you could review the code changes and test them.
If there are other functions that might be affected by the schema size limit, we could consider adding similar test cases for those as well.
Please double-check that the expected error type (MaxEncodedSchemaLimitExceeded) is accurate.
Thank you for your time and consideration. Please merge this pull request when you're satisfied with the changes.

Best,

Harish Karthick S

### DIFF
--- a/pallets/schema/src/tests.rs
+++ b/pallets/schema/src/tests.rs
@@ -315,47 +315,29 @@ fn test_schema_lookup() {
 	});
 }
 
-#[cfg(test)]
-mod tests {
-	// This module contains unit tests for the `pallet/schema` module.
-	// Unit tests are isolated tests that focus on the functionality of a single module
-	// and are typically run during development and testing to ensure the module
-	// behaves as expected.
-	// The `#[cfg(test)]` attribute ensures this module is only compiled when the
-	// `test` feature is enabled.
-	use super::*;
-	use mock::MockStorage; // Assuming you have a mock for storage
+#[test]
+fn test_max_encoded_schema_limit_exceeded_create() {
+	// Arrange
+	let mut mock_storage = MockStorage::new();
 
-	#[test]
-	fn test_max_encoded_schema_limit_exceeded_create() {
-		// Arrange
-		let mut mock_storage = MockStorage::new();
-		const MAX_SCHEMA_SIZE: usize = 1024; // Replace with the actual value
+	// Act
+	let large_schema_data = vec![0u8; <Test as frame_system::Config>::MaxEncodedSchemaLength + 1];
+	let result =
+		Schema::create(Origin::signed(DID_00), large_schema_data.clone(), Default::default());
 
-		// Act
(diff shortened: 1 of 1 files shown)


pallets/schema/src/tests.rs
```diff
@@ -314,3 +314,48 @@ fn test_schema_lookup() {
 		}
 	});
 }
+
+#[cfg(test)]
```
smohan-dw: It is better to extend the existing framework with new tests than create a mod. The #[cfg(test)] is activated with #[test] macro. Also, rather than passing a custom MAX_SCHEMA_SIZE, use the MaxEncodedSchemaLength value from mock.

HarishKarthickS: Sir, I have made [Changes according to the review](https://github.com/dhiway/cord/pull/435/commits/3387095bd40868e0a41ae829290f22206c8822e4)

pallets/schema/src/tests.rs
```diff
@@ -314,3 +314,30 @@ fn test_schema_lookup() {
...
 	});
 }
+
+#[test]
+fn test_max_encoded_schema_limit_exceeded_create() {
+	// Arrange
+	let mut mock_storage = MockStorage::new();
+
+	// Act
+	let large_schema_data = vec![0u8; <Test as frame_system::Config>::MaxEncodedSchemaLength + 1];
```
vatsa287, line 324: @HarishKarthickS I don't think this can be done, as the input is compiled to be of type `BoundedVec` and not just `Vec`, where `MaxEncodedSchemaLength + 1` is impossible due to below,
`pub type InputSchemaOf<T> = BoundedVec<u8, <T as Config>::MaxEncodedSchemaLength>;
`
but it is possible for `MaxEncodedSchemaLength - 1` to take place with a `BoundedVec` since it will be below threshold input set limit.

VedantKhairnar:
@HarishKarthickS 
Are you working on this PR?

PR #438, opened 2024-04-27
Updated ReadMe: Improved docker command to start running a container with custom name

Exar04:
To give custom name to a container the `--name <CustomeName>` tag should be use directly after docker run


vatsa287 (approved):
LGTM

PR #439, opened 2024-05-02
Asset: Add tests for AuthorizationNotFound

Pratik-050:
Details:
Created a new function asset_create_unauthorized_operation to generate an asset with invalid authorization.

### DIFF
--- a/pallets/asset/src/tests.rs
+++ b/pallets/asset/src/tests.rs
@@ -2,7 +2,7 @@ use super::*;
 use crate::mock::*;
 use codec::Encode;
 use cord_utilities::mock::{mock_origin::DoubleOrigin, SubjectId};
-use frame_support::{assert_ok, BoundedVec};
+use frame_support::{assert_err, assert_ok, BoundedVec};
 use frame_system::RawOrigin;
 use pallet_chain_space::{SpaceCodeOf, SpaceIdOf};
 use sp_runtime::{traits::Hash, AccountId32};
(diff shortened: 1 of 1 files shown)


Pratik-050:
Hello @vatsa287 I've made a PR for testing unauthorized operations, please review it.

Pratik-050:
Hello @vatsa287 @amarts can you please review this PR, and let me know if any changes are required.

pallets/asset/src/tests.rs
```diff
@@ -495,3 +495,62 @@ fn asset_vc_status_change_should_succeed() {
...
+		asset_value,
+		asset_tag,
+		asset_meta,
+	};
+
+	let digest = <Test as frame_system::Config>::Hashing::hash(&entry.encode()[..]);
+
+	new_test_ext().execute_with(|| {
+		// Attempt to create an asset without proper authorization
+		let result = Asset::create(
```
vatsa287, line 538: Instead use `assert_err` where the first arg is to create a `Asset` with invalid `authorisation_id`, second arg is the error expected here it is `AuthorizationNotFound`. 
As in,
```
		assert_err!(
			Asset::create(
				DoubleOrigin(author.clone(), creator.clone()).into(),
				entry,
				digest,
				invalid_authorization_id,
			),
			pallet_chain_space::Error::<Test>::AuthorizationNotFound
		);
```
You might have to import above error enum.

vatsa287: Instead use assert_err where the first arg is to create a Asset with invalid authorisation_id, second arg is the error expected here it is AuthorizationNotFound.
As in,

		assert_err!(
			Asset::create(
				DoubleOrigin(author.clone(), creator.clone()).into(),
				entry,
				digest,
				invalid_authorization_id,
			),
			pallet_chain_space::Error::<Test>::AuthorizationNotFound
		);
You might have to import above error enum.
It is good practice if you can actually `assert` with exact defined error enum from the pallet instead of a string.

pallets/asset/src/tests.rs
```diff
@@ -495,3 +495,62 @@ fn asset_vc_status_change_should_succeed() {
 		));
 	});
 }
+
+#[test]
+fn asset_create_unauthorized_operation() {
```
vatsa287: Change method name to `asset_functionalities_with_invalid_authorization_should_fail`

pallets/asset/src/tests.rs
```diff
@@ -495,3 +495,62 @@ fn asset_vc_status_change_should_succeed() {
...
+
+	new_test_ext().execute_with(|| {
+		// Attempt to create an asset without proper authorization
+		let result = Asset::create(
+			DoubleOrigin(author.clone(), creator.clone()).into(),
+			entry,
+			digest,
+			invalid_authorization_id,
+		);
+
```
vatsa287: Add similar tests in this test method for all asset functionalities, here you have added for `create` similarly for all defined methods in `Asset` or all places where chain-space authorisation is required.

vatsa287 (changes requested):
This test is not technically part of `Asset` but for `Chain-Space`. I would take this in as `chain-space` auth is required as part of this pallet as well.

pallets/asset/src/tests.rs
Pratik-050: added this accordingly. No need to import the enum.

pallets/asset/src/tests.rs
Pratik-050: I guess here no other function requires chain-space authorization.

vatsa287 (changes requested):
@Pratik-050 Please check where else this error code would be used. Not just in `create`.

PR #428, opened 2024-04-11
Remove deprecated getter() calls from all pallets

vatsa287:
- Remove deprecated `getter()`storage calls from all pallets.

Updates: #426

Issue #426, opened 2024-04-10
[Deprecation] remove storage getters (pallet::getter) usage from all pallets


smohan-dw:
This issue tracks the removal of pallet::getter in all pallets, as per https://github.com/paritytech/polkadot-sdk/issues/223. The preferred syntax is Key::<T>::get(). 

- [x] asset
- [x] authority-membership
- [x] chain-space
- [x] did
- [x] did-name
- [x] identity
- [x] membership
- [x] network-membership
- [x] network-score
- [x] node-authorization
- [x] offences
- [x] schema
- [x] statement

vatsa287:
Changes yet to be made on,

- [ ] test-utils/runtime
- [x] runtime/authorities

Will open a new PR for it.

kartikaysaxena:
Can I collaborate on this @vatsa287 ?

vatsa287:
> Can I collaborate on this @vatsa287 ?

@kartikaysaxena sure!
I have added a TODO in test-utils/runtime where the change needs to be made. 
Currently the issue is type resolution of generic types is not happening.

PR #440, opened 2024-05-02
Remove deprecated getter

kartikaysaxena:
Related #426 and #428


kartikaysaxena:
PTAL @vatsa287 is this approach correct?

kartikaysaxena:
I now get the issue, the function has some unsatisfied trait bounds. Would fix this.

Issue #383, opened 2024-03-12
[C4GT] Asset: Add tests for status_change of asset


vatsa287:
## Description
Subtask under : https://github.com/cord-network/community/issues/7

New testcase can be tested under `cargo test -p pallet-asset` after adding it in the code.

## Goals

- [ ] Add tests for status_change of asset unit case.

## Expected Outcome
- [ ] Test should assert for status_change of asset with passing of assertion.
       Asset status should be changed. Example from `active` to `inactive` or `expired`.
       
## Acceptance Criteria
NA

## Implementation Details
Look at other test cases, and add a test case for the same.
Ex: check method `asset_create_should_succeed` which validates creation of asset.

## Mockups / Wireframes
NA

---- 

## Product Name
CORD

## Organization Name
Dhiway

## Domain
Blockchain

## Tech Skills Needed
Rust

## Mentor(s)
@amarts @vatsa287 

## Complexity
[Medium]

## Category
[Test]

## Sub Category
[Beginner friendly]

Soham-27:
Hello @vatsa287, I can take on this issue with confidence. Please assign it to me and I'll ensure that it gets resolved effectively.

Soham-27:
hello @vatsa287 ,I have raised PR for this issue .

zeel991:
> hello @vatsa287 ,I have raised PR for this issue .
> 

Hello Soham , link your PR with the respective issue , so everyone knows that someone is already solving the issue

Soham-27:
Is this correct?

amarts:
We are moving away from the 'asset' pallet described here to pallet/assets instead of what exists now:

https://paritytech.github.io/polkadot-sdk/master/pallet_assets/index.html

Hence, closing this issue.  

Refer to https://github.com/dhiway/cord/blob/develop/docs/cord-features.md to get more idea about what features you can contribute on.

PR #445, opened 2024-05-08
fix for #383

Ektaroy:


vatsa287 (changes requested):
No need to add new files.
There is a file called `tests.rs` in `pallets/asset/`.
Please add the test case into that file.

PR #446, opened 2024-05-09
Benchmark(pallets, runtime): 2 * 8GB

vatsa287:
Benchmarking CORD pallet & runtime with 2(AMD EPYC 7B12) x 8 GB Machine (linux).

PR #447, opened 2024-05-09
Benchmark(pallets, runtime): 2 * 16GB

vatsa287:
Benchmarking CORD pallet & runtime with 2(AMD EPYC 7B13) x 16 GB Machine (linux).


vatsa287:
@amarts Now both benchmarks are done on AMD EPYC based machines.

amarts:
Let us go with this, instead of #446.

amarts (approved):

PR #450, opened 2024-05-13
Add custom metadata storage

vatsa287:
This is a draft implementation of having a custom `metadata` of `MaxMetaPairLength` in asset, but in general this can be replicated across the chain.
One of the current use-case would be to store `created_at: UTC of old_chain_block` after migration so apps can access this value instead of newly created block time.

### How it is implemented?
- For demonstration have made this to `asset` specific.
- Create a storage named `AssetMeta` to store metadata of storage. It allows to custom k:v of `MaxMetaPairLength` currently set to 5, where the `identifier` of the asset is the first key, and second key is a custom type. 
- When lookup'ed on using the identifier we get a list of all the `k:v` of the pallet metadata, but when used with a custom `key` it produces a result with the `custom value`.

Ex:
- Create a metapairs on the SDK, with `string` type `k:v` which has custom keys and values attached to the identifier of the asset.
```typescript
    const metaPairs: [string, string][] = [
      ["key1", "value1"],
      ["key2", "value2"],
      ["key3", "value3"],
      ["key4", "value4"],
      ["key5", "value5"],
    ];

    const tx = api.tx.asset.issue(
      assetEntry.entry,
      assetEntry.digest,
      authorizationId,
      metaPairs
    )
```

- Chain-state after execution with all `k:v` seen,
Screenshot 2024-05-13 at 3 06 54 PM

- Chain-state after execution with a particular key query,
Screenshot 2024-05-13 at 3 07 13 PM

### DIFF
--- a/pallets/asset/src/benchmarking.rs
+++ b/pallets/asset/src/benchmarking.rs
@@ -175,7 +175,7 @@ benchmarks! {
 			pallet_chain_space::Pallet::<T>::approve(chain_space_origin, space_id, capacity).expect("Approval should not fail.");
 			Pallet::<T>::create(origin.clone(), entry, digest, authorization_id.clone())?;
 
-		}: _<T::RuntimeOrigin>(origin, issue_entry, issue_entry_digest, authorization_id)
+		}: _<T::RuntimeOrigin>(origin, issue_entry, issue_entry_digest, authorization_id, None)
 		verify {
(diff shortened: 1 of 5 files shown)


vatsa287:
@amarts @smohan-dw @NiranjanAP @atreyasuraj ^^

PR #451, opened 2024-05-16
Witness: Add new pallet witness

vatsa287:
Add a new pallet named `witness` which has two methods,
- `Create`: Create a witness requirement for the given document by specifying the required `witness_count`
- `Witness`: Witnesses sign the document with their respective DIDs.


amarts (approved):
this is getting into `dmp_2024` branch.

PR #453, opened 2024-05-22
scripts [Minor]: Remove collective from list of pallets

vatsa287:
Since now the pallet `collective` has been removed from the CORDs pallet list, but exists in runtime more as a dependency.
Remove `PALLET_COLLECTIVE` from the list of pallets for which the weight must be generated.

### DIFF
--- a/scripts/run_benches_for_pallets.sh
+++ b/scripts/run_benches_for_pallets.sh
@@ -60,7 +60,6 @@ CORD=./target/production/cord
 # Manually exclude some pallets.
 PALLETS=(
   "pallet_chain_space"
-  "pallet_collective"
   "pallet_did"
   "pallet_did_name"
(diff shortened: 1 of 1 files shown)


smohan-dw (approved):
LGTM

smohan-dw (commented):
pub type TechnicalCommittee = pallet_collective<Instance2>;
pallet_collective is not removed from runtime. We should continue to benchmark this pallet.

smohan-dw (changes requested):
pub type TechnicalCommittee = pallet_collective<Instance2>; 
we should continue to benchmark this pallet as it is part of the runtime

vatsa287:
Ack. Will close this now.

vatsa287:
Reopening as it is a valid PR.

PR #465, opened 2024-06-11
Asset: added test cases for asset and VC over issuance

atreyasuraj:


pallets/asset/src/tests.rs
```diff
@@ -138,29 +138,6 @@ fn asset_issue_should_succeed() {
 
 	let issue_entry_digest =
 		<Test as frame_system::Config>::Hashing::hash(&[&issue_entry.encode()[..]].concat()[..]);
-
-	new_test_ext().execute_with(|| {
```
vatsa287: nit: Maybe this has been removed by mistake.

pallets/asset/src/tests.rs
```diff
@@ -874,3 +852,185 @@ fn changing_status_of_vc_asset_with_same_status_should_fail() {
...
+				issue_entry_digest,
+				authorization_id
+			),
+			Error::<Test>::OverIssuanceLimit
+		);
+	});
+}
+
+#[test]
+fn asset_vc_status_change_should_not_succeed() {
```
vatsa287, line 936: Please change the test-case name since this already exists.

pallets/asset/src/tests.rs
```diff
@@ -874,3 +852,185 @@ fn changing_status_of_vc_asset_with_same_status_should_fail() {
...
+		);
+
+		assert_err!(
+			Asset::vc_status_change(
+				DoubleOrigin(author.clone(), creator.clone()).into(),
+				asset_id.clone(),
+				Some(instance_id.clone()),
+				AssetStatusOf::INACTIVE
+			),
+			Error::<Test>::AssetInstanceNotFound
```
vatsa287, line 1033: nit: Last test may not be required.

vatsa287 (dismissed):
LGTM
nit: Would be right if the test case name be related to over issuance, as there might multiple ways issue, status_change may fail.

atreyasuraj:
> LGTM nit: Would be right if the test case name be related to over issuance, as there might multiple ways issue, status_change may fail.



> LGTM nit: Would be right if the test case name be related to over issuance, as there might multiple ways issue, status_change may fail.

Sure, fixed!

amarts (approved):

PR #487, opened 2024-07-15
statement: Add selective data keypairs as storage

vatsa287:
This PR allows to have `selective data` as storage in the chain-state.
This feature allows to have individual fields of digest in a document, with a maximum of 25 k:v's.
The key is custom, and value must be a digest of 32 length.

This implementation allows to check for availability of individual fields of a document has been anchored or not instead of whole document itself. This is crucial when verification is done using a non-digital copy as a reference digest.

Updates:

- `register`: Optional argument `selective_data` to add.
- `update`: Optional argument `selective_data` to update existing `selective_data`.
- `update_selective_data`: Update just the `selective_data`.

#
- `remove_selective_data`: Remove existing `selective_data` to free up space entry usage.

### DIFF
--- a/runtimes/braid/src/lib.rs
+++ b/runtimes/braid/src/lib.rs
@@ -742,7 +742,7 @@ impl pallet_chain_space::Config for Runtime {
 parameter_types! {
 	pub const MaxDigestsPerBatch: u16 = 1_000;
 	pub const MaxRemoveEntries: u16 = 1_000;
-	pub const MaxSelectiveDataKeyLength: u32 = 100;
+	pub const MaxSelectiveDataKeyLength: u32 = 64;
 	pub const MaxSelectiveDataEntries: u32 = 25;
 }
 
--- a/runtimes/braid/src/lib.rs
+++ b/runtimes/braid/src/lib.rs
@@ -742,6 +742,8 @@ impl pallet_chain_space::Config for Runtime {
 parameter_types! {
 	pub const MaxDigestsPerBatch: u16 = 1_000;
 	pub const MaxRemoveEntries: u16 = 1_000;
+	pub const MaxSelectiveDataKeyLength: u32 = 100;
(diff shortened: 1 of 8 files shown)


runtimes/braid/src/lib.rs
```diff
@@ -742,6 +742,8 @@ impl pallet_chain_space::Config for Runtime {
 parameter_types! {
 	pub const MaxDigestsPerBatch: u16 = 1_000;
 	pub const MaxRemoveEntries: u16 = 1_000;
+	pub const MaxSelectiveDataKeyLength: u32 = 100;
```
amarts: Can we keep the length to 64? I guess it should be enough.

runtimes/weave/src/lib.rs
```diff
@@ -885,6 +885,8 @@ impl pallet_chain_space::Config for Runtime {
 parameter_types! {
 	pub const MaxDigestsPerBatch: u16 = 1_000;
 	pub const MaxRemoveEntries: u16 = 1_000;
+	pub const MaxSelectiveDataKeyLength: u32 = 100;
```
amarts: Same here, 64 would be good is what I think. and if its 64, should we keep this as u32 or u16/u8?

runtimes/braid/src/lib.rs
vatsa287: Ack. It will be enough.

runtimes/weave/src/lib.rs
vatsa287: Technically u8/u16 is enough. But some traits are not implemented for those types which is used in storage. So have used u32. Have to check.

runtimes/braid/src/lib.rs
vatsa287: Updated.

vatsa287:
PTAL @smohan-dw @amarts @NiranjanAP

amarts (approved):

PR #490, opened 2024-07-24
feat: add pallet:UseOnce  for consumable statements

Harsh1s:


smohan-dw:
What features distinguish this pallet from the current statement pallet?

PR #493, opened 2024-08-01
run bootstrap command on docker image

Rithwik0480:
- Modified bootstrap.sh to use Docker image for obtaining the spec file and generating node keys instead of compiling CORD directly.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -88,7 +88,7 @@ echo "
 " >>$ACCOUNTS_FILE
 echo "" >>$ACCOUNTS_FILE
 
-cord='./target/release/cord'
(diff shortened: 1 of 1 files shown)


scripts/bootstrap.sh
```diff
@@ -88,7 +88,7 @@ echo "
 " >>$ACCOUNTS_FILE
 echo "" >>$ACCOUNTS_FILE
 
-cord='./target/release/cord'
+cord="sudo docker run -v $OUTPUT_DIR:$OUTPUT_DIR dhiway/cord:develop"
```
amarts, line 91: can you have the check for building OUTPUT_DIR as the absolute path, instead of relative path?

PR #495, opened 2024-08-01
DeDir: Add new pallet `Decentralized Directory`

vatsa287:
A `v1` implementation of DeDir - Decentralised Directory.

- [x] Basic Functionalities - `create-registry`, `create-registry-entry`, `registry-entry-state-change`.
- [x] Change states to be in `enum` format. Custom states will be from `attributes`.
- [x] Identifier creation/management.
- [x] Delegation/ Registry-Ownership management.

Abandoned. Do not merge.

### DIFF
--- a/pallets/dedir/src/lib.rs
+++ b/pallets/dedir/src/lib.rs
@@ -1,0 +1,456 @@
+// This file is part of CORD – https://cord.network
+
+// Copyright (C) Dhiway Networks Pvt. Ltd.
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+// CORD is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// CORD is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with CORD. If not, see <https://www.gnu.org/licenses/>.
+//
+
+//! # Decentralized Directory Pallet - DeDir
+//!
+//!
+//! ## Overview
+//!
+//! The Decentralized Directory Pallet (DeDir) aims to implement a decentralized
+//! version of a registry. Enabling creation, updation of registries in a distributed
+//! decentralized manner. Thereby enabling trust and transperency utilizing CORD blockchain.
+//!
+//! ## Interface
+//!
+//! ### Dispatchable Functions
+//!
+//! * `create_registry` - Create a registry, with states supported and entry types.
+//! * `create_registry_entry` - Create a registry entry for the created registry.
+//! * `registry_entry_status_change` - Change the status of the registry entry.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+mod types;
+
+use frame_support::{
+	ensure,
+	pallet_prelude::DispatchResult,
+	traits::{Get, StorageVersion},
+	BoundedVec,
+};
+pub use pallet::*;
+use sp_std::{prelude::*, str};
+
+pub use frame_system::WeightInfo;
+pub use types::{Data, Entry, Registry, RegistryEntry, RegistrySupportedTypeOf};
+
+#[frame_support::pallet]
+pub mod pallet {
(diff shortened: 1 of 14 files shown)


amarts (dismissed):
LGTM

Is there any demo script planned to be implemented for the same?

vatsa287:
> LGTM
> 
> Is there any demo script planned to be implemented for the same?

@amarts 
Sure it will be implemented.
It would be great if you can merge [this PR](https://github.com/dhiway/cord.js/pull/231), so I can push on top of this.

pallets/dedir/src/lib.rs
```diff
@@ -0,0 +1,456 @@
...
+		///
+		/// # Example
+		/// ```
+		/// create_registry(origin, registry_id, attrs, additional_states)?;
+		/// ```
+		#[pallet::call_index(0)]
+		#[pallet::weight({0})]
+		pub fn create_registry(
+			origin: OriginFor<T>,
+			registry_id: RegistryIdOf<T>,
```
smohan-dw: Registry ID is an identifier that should be returned after a successful operation.

pallets/dedir/src/lib.rs
```diff
@@ -0,0 +1,456 @@
...
+			attrs: Vec<(RegistryKeyIdOf<T>, RegistrySupportedTypeOf)>,
```
smohan-dw: The registry entry attributes should also help to identify the lookup fields that should be used for creating entry identifiers

pallets/dedir/src/lib.rs
```diff
@@ -0,0 +1,456 @@
...
+			// TODO: Use Runtime Constant for upper bound
+			additional_states: Option<BoundedVec<BoundedVec<u8, ConstU32<32>>, ConstU32<10>>>,
```
smohan-dw: why can't this be enums? also why can't this be part of attrs?

pallets/dedir/src/lib.rs
```diff
@@ -0,0 +1,456 @@
...
+			}
+
+			/* Ensure new-states does not overflow */
+			if let Some(states) = additional_states {
+				if registry.add_supported_states(states).is_err() {
+					return Err(Error::<T>::AddingStateFailed.into());
+				}
+			}
+
+			Registries::<T>::insert(&registry_id, registry);
```
smohan-dw, line 327: registry ownership?

pallets/dedir/src/lib.rs
```diff
@@ -0,0 +1,456 @@
...
+		/// ```
+		/// create_registry_entry(origin, registry_id, registry_entry_id,
+		///     digest, attrs)?;
+		/// ```
+		#[pallet::call_index(1)]
+		#[pallet::weight({0})]
+		pub fn create_registry_entry(
+			origin: OriginFor<T>,
+			registry_id: RegistryIdOf<T>,
+			registry_entry_id: RegistryEntryIdOf<T>,
```
smohan-dw: again registry_entry_id is the result of the operation

pallets/dedir/src/lib.rs
```diff
@@ -0,0 +1,456 @@
...
+			digest: EntryHashOf<T>,
+			attrs: Option<Vec<(RegistryKeyIdOf<T>, Data)>>,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
```
smohan-dw: need to use this to check the registry ownership or delegation

smohan-dw (changes requested):

pallets/dedir/src/lib.rs
vatsa287: With the discussion with @amarts  previously if the identifier is available then the registry_entry attributes would be fetch able.

pallets/dedir/src/lib.rs
vatsa287: Ack.
As per newer design the states supported shall not be dynamic as implemented above but fixed to `Active`, `Revoked`, `Draft` which will become enum type.
So any custom state can be part of attrs itself.

pallets/dedir/src/lib.rs
vatsa287: As per earlier conversation `identifier` creation was passed onto app side. 
Would move the logic here if required.

pallets/dedir/src/lib.rs
vatsa287: Delegation/Ownership is not implemented in this version. Will do with next commit.

amarts:
done with #504

PR #499, opened 2024-08-06
Node-Authorization : Test for `PeerIdTooLong`

### LINKED ISSUE #305: [C4GT] Node-Authorization: Add tests for `PeerIdTooLong`
## Description

New testcase can be tested under `cargo test -p pallet-node-authorization` after adding it in the code.

## Goals

- [ ] Add tests for `PeerIdTooLong` for pallet/node-authorization

## Expected Outcome
- [ ] Test should assert for `PeerIdTooLong` being returned properly in all the possible calls.
       
## Acceptance Criteria
NA

## Implementation Details
Ex: check method `add_well_known_node_works` which validates error code `NodeIdTooLong`

## Mockups / Wireframes
NA

---- 

## Product Name
CORD

## Organization Name
Dhiway

## Domain
Blockchain

## Tech Skills Needed
Rust

## Mentor(s)
@amarts 

## Complexity
[Medium]

## Category
[Test]

## Sub Category
[Beginner friendly]

zeel991:
- Fixes https://github.com/dhiway/cord/issues/305
- This PR will check for all the conditions where `PeerIdTooLong` could be returned

### DIFF
--- a/pallets/node-authorization/src/tests.rs
+++ b/pallets/node-authorization/src/tests.rs
@@ -461,8 +461,8 @@ fn test_generate_peer_id_invalid_utf8() {
 }
 
 #[test]
-fn peer_id_too_long_test() -> Result<(),Error<Test>>{
-    new_test_ext().execute_with(|| {
+fn peer_id_too_long_test() -> Result<(), Error<Test>> {
+	new_test_ext().execute_with(|| {
         const max_peer_id_length: usize = 128;
         let testing = "nodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeidnodeid";
 
--- a/pallets/node-authorization/src/tests.rs
+++ b/pallets/node-authorization/src/tests.rs
@@ -463,11 +463,11 @@ fn test_generate_peer_id_invalid_utf8() {
 #[test]
 fn peer_id_too_long_test() {
 	new_test_ext().execute_with(|| {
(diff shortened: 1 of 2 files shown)


zeel991:
Hello @vatsa287 , The `PeerId` is created when `node_id` is converted into **byte vector** which is later passed in `generate_peer_id` function which encodes it in **UTF8** and decodes it in **BASE58** and the decoded variable is wrapped in a struct . 
If I have understood it correctly , can you please give me a hint as to how to generate a peer_id greater than `MAX_PEER_ID_LENGTH` but also keep the node_id less than the `MAX_NODE_ID_LENGTH `

vatsa287:
@zeel991 
From my understanding it is not possible to test it by calling the method itself. Instead test for `generate_peer_id` itself. Check if the produced output is greater than allowed length for a peer id.
The upper bound constants are defined in `mock.rs` for test usage.

pallets/node-authorization/src/tests.rs
```diff
@@ -459,3 +459,15 @@ fn test_generate_peer_id_invalid_utf8() {
 	let invalid_node_id: NodeId = vec![0xFF, 0xFE, 0xFD];
 	assert_err!(NodeAuthorization::generate_peer_id(&invalid_node_id), Error::<Test>::InvalidUtf8);
 }
+
```
vatsa287, line 462: Could you try doing this below

```
#[test]
fn peer_id_too_long_test() {
	new_test_ext().execute_with(|| {
        let long_node_id = "nodeid".repeat(32); 

        let node_identity = test_node(&long_node_id);
        let peer_id = NodeAuthorization::generate_peer_id(&node_identity).unwrap();

        assert_eq!(
            test_peer_id_length(peer_id), 
            Err(Error::<Test>::PeerIdTooLong)
        );
    })
}
```

zeel991: Done , To avoid `mismatched type` error , I changed the function `test_peer_id_length` to return a Result in mock.rs

zeel991:
@vatsa287 The test has passed but the build action has failed due to some other reason

vatsa287 (approved):
LGTM

zeel991:
@vatsa287 , I would like to work on some more issues , do you have any specific issues for me  ?

PR #500, opened 2024-08-11
Runtime Re-Org - Relay, Appchain & Standalone

smohan-dw:
ReOrg Details

New Runtimes
- CORD Loom - Relaychain using Polkadot relay protocol
- CORD Weave - Appchain using Substrate and Cumulus

Changes to Standalone Runtime
- Standalone renamed to CORD Braid
- Supports two runtimes
   - Braid Base - Standalone without Governance
   - Braid Plus - Standalone with Governance
- Removed network membership requirement as the accounts are now using core units
- Had to rename Statement pallet to DId-Token due to upstream conflict (todo - need to finalise the pallet name)


amarts:
Just had a glance.  340 files changed, and may take some time to review. I will block sometime tomorrow for this.

about CI, would prefer to offload some work on this.

- Is the rustfmt version you are having, and the one in CI different? should we update CI?
- build is good, clippy will check tomorrow.

Can we do a discussion on this during community meeting?

smohan-dw:
> Just had a glance. 340 files changed, and may take some time to review. I will block sometime tomorrow for this.
> 
> about CI, would prefer to offload some work on this.
> 
> * Is the rustfmt version you are having, and the one in CI different? should we update CI?
> * build is good, clippy will check tomorrow.

will check that before removing the draft status
> 
> Can we do a discussion on this during community meeting?

good idea. we can do that.

amarts (dismissed):
I see some conflicts. Will pick it up as a follow-up.

PR #502, opened 2024-08-23
dedir: Add new pallet for `Decentralised Directories` with `blob`

vatsa287:
Updates: #503

Abandoned. Do not merge.

### DIFF
--- a/pallets/dedir/src/types.rs
+++ b/pallets/dedir/src/types.rs
@@ -1,0 +1,99 @@
+// This file is part of CORD – https://cord.network
+
+// Copyright (C) Parity Technologies (UK) Ltd.
+// Copyright (C) Dhiway Networks Pvt. Ltd.
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Adapted to meet the requirements of the CORD project.
+
+// CORD is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// CORD is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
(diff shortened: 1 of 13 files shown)


pallets/dedir/src/types.rs
```diff
@@ -0,0 +1,99 @@
...
+/// The `Permissions` enum defines the levels of access control available for an account within a
+/// registry.
+///
+/// - `DELEGATE`: Grants permission to manage registry entries.
+/// - `ADMIN`: Extends `DELEGATE` permissions, allowing the management of delegates in addition to
+///   managing registry entries.
+/// - `OWNER`: The creator or owner of the registry. This permission level encompasses the full
+///   range of management capabilities, including the permissions of both `DELEGATE` and `ADMIN`.
+#[derive(Encode, Decode, MaxEncodedLen, Clone, RuntimeDebug, PartialEq, Eq, TypeInfo)]
+pub enum Permissions {
```
smohan-dw: Permission as Enums will not help to create composite permissions. A delegate can have multiple roles associated with a registry. Please refer the chain-space permissioning system.

pallets/dedir/src/types.rs
```diff
@@ -0,0 +1,99 @@
...
+//! DeDir pallet types.
+
+use codec::{Decode, Encode, MaxEncodedLen};
+use scale_info::TypeInfo;
+
+use sp_runtime::RuntimeDebug;
+use sp_std::prelude::*;
+
+#[derive(Encode, Decode, MaxEncodedLen, Clone, RuntimeDebug, PartialEq, Eq, TypeInfo)]
+pub enum RegistrySupportedStateOf {
```
smohan-dw, line 32: Do we need a DRAFT state at the chain level? Without that this can be a bool filed representing the status

amarts: This thought process came when we asked the question if chain alone is used as 'store' also, how to differentiate between something which is not ready to be 'lookedup' (or queried), and the one record which is ready to do the lookup.

Having these states covers most of the usecases is what we thought.

pallets/dedir/src/types.rs
vatsa287: Ack,
Have updated the flow to use bitwise and handle composite permissions.

amarts:
done with #504

PR #514, opened 2024-10-17
schema-accounts: Add pallet for accounts based schema management

vatsa287:
This PR,
- adds a new pallet called `schema-accounts` for account based schema accounts.
- removes dependency of `chain-space`. And does not require a `authId` for creation.
- identifier updates.
- update runtime to `0.9.3-1` or `9310`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -540,7 +540,7 @@ dependencies = [
 
 [[package]]
 name = "authority-membership"
-version = "0.9.4"
+version = "0.9.3-1"
 dependencies = [
  "frame-benchmarking",
  "frame-support",
@@ -1234,7 +1234,7 @@ checksum = "6245d59a3e82a7fc217c5828a6692dbc6dfb63a0c8c90495621f7b9d79704a0e"
 
 [[package]]
 name = "cord-braid-runtime"
-version = "0.9.4"
+version = "0.9.3-1"
 dependencies = [
  "authority-membership",
  "cord-braid-runtime-constants",
@@ -1318,7 +1318,7 @@ dependencies = [
 
 [[package]]
 name = "cord-braid-runtime-constants"
-version = "0.9.4"
+version = "0.9.3-1"
 dependencies = [
  "cord-primitives",
  "cord-runtime-common",
@@ -1348,7 +1348,7 @@ dependencies = [
 
 [[package]]
 name = "cord-identifier"
-version = "0.9.4"
+version = "0.9.3-1"
 dependencies = [
  "blake2-rfc",
  "bs58 0.5.1",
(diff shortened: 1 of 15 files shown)


amarts (commented):
LGTM.

Few questions, 
* is 9400 better, or 9310 better as newer version?
* is 'schema-accounts' a good name? or something else?
* should schema identifier be different for account based and did based? or should it be using same formula regardless of creator, in which case, we shouldn't use creator identity in schema identifier creation.

vatsa287:
> #is 9400 better, or 9310 better as newer version?

I believe `9310` can be made, and next upcoming PR can have it updated to `9400` when it integrates account based schema with `registries` pallet. I am not sure if both `9300` and `9310` stands for `0.9.3` or it differs.

> is 'schema-accounts' a good name? or something else?

I thought it tells what it is based on, I had something like `schema-registries` on mind as well since it created to work with it. If you find that good or you have something else in mind, please suggest.

> should schema identifier be different for account based and did based? or should it be using same formula regardless of creator, in which case, we shouldn't use creator identity in schema identifier creation.

In `schema` pallet it was based on `creator`, `space_id` & `schema input`. Here it is just `creator` & `schema input`. Please tell me if you want to remove `creator` as part of `identifier`. I believe it is not required, as `schema` here can be public as well. 

^ @amarts

amarts:
> > #is 9400 better, or 9310 better as newer version?
> 
> I believe `9310` can be made, and next upcoming PR can have it updated to `9400` when it integrates account based schema with `registries` pallet. I am not sure if both `9300` and `9310` stands for `0.9.3` or it differs.
> 

9310 is better in that case. 

> > is 'schema-accounts' a good name? or something else?
> 
> I thought it tells what it is based on, I had something like `schema-registries` on mind as well since it created to work with it. If you find that good or you have something else in mind, please suggest.
>

OK for now. Lets take this up again.
 
> > should schema identifier be different for account based and did based? or should it be using same formula regardless of creator, in which case, we shouldn't use creator identity in schema identifier creation.
> 
> In `schema` pallet it was based on `creator`, `space_id` & `schema input`. Here it is just `creator` & `schema input`. Please tell me if you want to remove `creator` as part of `identifier`. I believe it is not required, as `schema` here can be public as well.
> 

The creator needs to be captured for the query, but need not be used for the schema identifier. That way, both pallets use the same logic for identifiers.

vatsa287:
@amarts 
Have updated the identifier logic by keeping only `schema_input` for identifier creation part.

[commit](https://github.com/dhiway/cord/pull/514/commits/dbfc4aa7c6218b0e2564faac4d3dd83e910ac6d1) for ref.

amarts (approved):

PR #516, opened 2024-10-18
Integrate `schema-accounts` with `registries` & `entries`

vatsa287:
This PR,
- Integrates the account based schema-management pallet `schema-accounts` into `registries` & `entries`.
- Updates the runtime spec to `9400` or cord package to `0.9.4`.


pallets/registries/src/lib.rs
```diff
@@ -607,6 +609,7 @@ pub mod pallet {
 					revoked: false,
 					archived: false,
 					digest,
+					schema_id,
```
vatsa287, line 612: @amarts Here do you see the need of storing the `schema_id` with the `registry` is associated with?

amarts: Yes. It would be good to store.

vatsa287: Ack.

amarts (approved):

PR #518, opened 2024-10-28
weave-runtime: reduce the existential deposit to 100 milli units

### LINKED ISSUE #517: weave-runtime: found an issue of not enough balance even though balance is 100 UNITS
when testing https://faucet.cord.network/ to get an account loaded with few units, realised that if an account has any number of UNITS less than 100, it was not able to make any transactions.

This should be allowed even though an account has minimum amount to handle bare minimum transactions weights.

amarts:
in a permission-less runtime, user's minimum account balance should be near 0.

fixes: #517

### DIFF
--- a/runtimes/weave/constants/src/lib.rs
+++ b/runtimes/weave/constants/src/lib.rs
@@ -25,13 +25,13 @@ pub mod weights;
 pub mod currency {
 	use cord_primitives::Balance;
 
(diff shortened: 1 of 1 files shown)


smohan-dw:
LGTM

vatsa287 (approved):
Will send similar changes to `braid` & `loom` runtime as well.

Issue #370, opened 2024-03-12
[C4GT] Asset: Add tests for `InvalidSignature`


vatsa287:
## Description
Subtask under : https://github.com/cord-network/community/issues/7

New testcase can be tested under `cargo test -p pallet-asset` after adding it in the code.

## Goals

- [ ] Add tests for `InvalidSignature` for pallet/asset

## Expected Outcome
- [ ] Test should assert for `InvalidSignature` being returned properly in all the possible calls.
- [ ] If there are more than one function returning this error code, all of them should be validated either as one test case or as multiple test cases.
       
## Acceptance Criteria
NA

## Implementation Details
Look at other test cases, and add a test case for the same.
Ex: check method `asset_create_should_succeed` which validates creation of asset.

## Mockups / Wireframes
NA

---- 

## Product Name
CORD

## Organization Name
Dhiway

## Domain
Blockchain

## Tech Skills Needed
Rust

## Mentor(s)
@amarts @vatsa287 

## Complexity
[Low]

## Category
[Test]

## Sub Category
[Beginner friendly]

Aloneking789:
I have made a pull request Please merge it

vatsa287:
@Aloneking789 Welcome! I see multiple PRs being raised without being an assignee.
It would be really appreciated if you could go through https://github.com/dhiway/cord/discussions/396.
Apart from this [PR](https://github.com/dhiway/cord/pull/400), I would be closing other PRs for the reasons mentioned in the above discussion.

trishanu-init:
@vatsa287 Please assign this issue to me.

VedantKhairnar:
Hello @trishanu-init 
Do you need any assistance with the issue or any progress you might want to share?
Thanks.

rohansen856:
Hello @vatsa287 @VedantKhairnar i see that this issue has no activity for quite a while. If it's alright, i would like to work on this issue. in that case please assign me this issue.

rohansen856:
@vatsa287, I have solved this issue and raised a PR. Thank you.

HasanZaigam:
Hi @vatsa287, this issue is currently active. If it is open, please assign me to this issue.

amarts:
We are moving away from the 'asset' pallet described here to pallet/assets instead of what exists now:

https://paritytech.github.io/polkadot-sdk/master/pallet_assets/index.html

Hence, closing this issue.  

[...]

PR #520, opened 2024-11-07
pallet-asset: Add new test case for InvalidAssetValue

rohansen856:
# Goal
This pr solves the issue: #370 , adding new test cases for checking invalid asset values in `pallets/asset/src/tests.rs`.

#summary 

### Context:

creator, author, and capacity establish the initial conditions.
The test then generates unique identifiers for both the "space" and "authorization" using hashing functions to mimic blockchain data integrity practices.
### Asset Entry with Invalid Value:

The entry variable represents an asset structure, with asset_value deliberately set to -10 to simulate an invalid value scenario.
This asset entry will be used to test the error handling in the Asset::create function.
Test Execution and Assertions:

`Space::create` and `Space::approve` are called to prepare the environment, ensuring the space is created and approved before attempting asset creation.
The core test assertion is `assert_err!`, which confirms that calling `Asset::create` with the invalid asset value returns the InvalidAssetValue error as expected.

- This structured approach not only validates error handling in the Asset::create function but also helps maintain the overall robustness of the system by preventing invalid data from being stored.

### DIFF
--- a/pallets/asset/src/tests.rs
+++ b/pallets/asset/src/tests.rs
@@ -1,3 +1,5 @@
+use core::{borrow::Borrow, u32};
+
 use super::*;
 use crate::mock::*;
 use codec::Encode;
@@ -2842,6 +2844,7 @@ fn asset_vc_status_change_with_wrong_asset_id_should_fail() {
 
 /// Test case for validating that asset creation fails when provided with invalid asset values.
 /// It covers both cases where the failure may result from an invalid `asset_value` or `asset_qty`.
+#[test]
 fn asset_create_should_fail_for_invalid_asset_value() {
 	// Define the creator and author identities, along with the capacity for the "space"
 	let creator = DID_00;
@@ -2881,8 +2884,8 @@ fn asset_create_should_fail_for_invalid_asset_value() {
 		let invalid_value_entry = AssetInputEntryOf::<Test> {
 			asset_desc: asset_desc.clone(),
 			asset_qty: 10, // Valid quantity
-			asset_type,
-			asset_value: -10, // Invalid asset value
+			asset_type: asset_type.clone(),
+			asset_value: u32::MAX+1, // Invalid asset value(u32 overflow)
 			asset_tag: asset_tag.clone(),
 			asset_meta: asset_meta.clone(),
 		};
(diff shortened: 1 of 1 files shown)


pallets/asset/src/tests.rs
```diff
@@ -2839,3 +2839,69 @@ fn asset_vc_status_change_with_wrong_asset_id_should_fail() {
 		);
 	});
 }
+
+/// Test case for validating that asset creation fails when provided with an invalid asset value or out of space.
+fn asset_create_should_fail_invalid_asset_value() {
```
vatsa287: nit: `asset_create_should_fail_for_invalid_asset_value()`

vatsa287 (changes requested):
I see the requirement here that Invalid Asset Value may get triggered due either from `asset_qty` or `asset_value`. 
Please check for both cases.

vatsa287:
Use below for passing the rust format & add new commit.
`cargo +nightly fmt --all`

pallets/asset/src/tests.rs
rohansen856: @vatsa287 thanks for the review. I will fix the naming issue and also add the `assert_qty` assertion error in next PRs. thank you.

vatsa287:
@rohansen856 While at it you can also check for `0` as it only supports unsigned numbers.

rohansen856:
@vatsa287 I have added test for the following cases:
`Overflow Values`: Ensures asset creation fails when asset_value or asset_qty exceeds the maximum limit for their respective types.
`Zero Values`: Validates that asset creation fails when either asset_value or asset_qty is set to zero, as only positive, non-zero values are supported.
`Negative Values`: Adds validation for negative values on asset_value and asset_qty, ensuring that asset creation correctly rejects negative entries. did this a bit out of the way by passing the values like this:
```rust
let negative_value_entry = AssetInputEntryOf::<Test> {
	asset_desc: asset_desc.clone(),
	asset_qty: 10, // Valid quantity
	asset_type: asset_type.clone(),
	asset_value: (0 - 1), // Invalid (negative) asset value
	asset_tag: asset_tag.clone(),
	asset_meta: asset_meta.clone(),
};
```
let me know if any other changes or edge case inclusions are necessary. thank you.

vatsa287:
@rohansen856 Negative check is not required since it is handled at by the type itself.

rohansen856:
> @rohansen856 Negative check is not required since it is handled at by the type itself.

thanks for the follow up. i will remove the negative check and keep just the `overflow` and `0` value error and make a pr.

PR #524, opened 2024-11-13
statement: Disallow delegate-2 to revoke/update delegate-1's statement

vatsa287:
`<Demo PR>, <Do Not Merge>`

Example:
1. Creator creates a chainspace.
2. Adds a delegate `delegate-1` with `ASSERT` permission.
3. Adds a delegate `delegate-2` with `ASSERT` permission.
4. Create a statement `statement-1` from `delegate-1`.
5. Create a statement `statement-2` from `delegate-2`.
6. Try revoking/edit `statement-1` from `delegate-2` or vice versa. ( Should this be allowed, currently this followed in DeDir too).

### DIFF
--- a/pallets/statement/src/tests.rs
+++ b/pallets/statement/src/tests.rs
@@ -31,6 +31,7 @@ pub fn generate_authorization_id<T: Config>(digest: &SpaceCodeOf<T>) -> Authoriz
 
 pub(crate) const DID_00: SubjectId = SubjectId(AccountId32::new([1u8; 32]));
 pub(crate) const DID_01: SubjectId = SubjectId(AccountId32::new([5u8; 32]));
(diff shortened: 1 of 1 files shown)


vatsa287:
@smohan-dw @amarts 
I needed your insights on above.

If this should be allowed we can something like below in `entries`/ `statement`.

1. Check if the origin has a admin authorisation of the `chainspace`/ `registry`.
2. Check if the origin is the creator, `creator` is stored during creation of the `statement`/ `entry`.
3. Allow only if any one of above is true, disallow else.

smohan-dw:
The delegates should be able to do any operations on the entries posted by other delegates. Only the registry admins should have that capability. Also make sure we are checking for a delegate permissions for revoke operation. Not all delegates have this permission. Go ahead and make the changes needed for this.

PR #526, opened 2024-11-19
entries: Restrict update operation only to Admin & Creator

vatsa287:
Restrict the permission for only `admin` and `creator` to update the entry.


amarts (approved):
LGTM

Issue #318, opened 2024-02-06
 [C4GT] Statement: Add tests for `PresentationDigestAlreadyAnchored`


vatsa287:
## Description
Subtask under : https://github.com/cord-network/community/issues/7

New testcase can be tested under `cargo test -p pallet-statement` after adding it in the code.

## Goals

- [ ] Add tests for `PresentationDigestAlreadyAnchored` for pallet/statement

## Expected Outcome
- [ ] Test should assert for `PresentationDigestAlreadyAnchored` being returned properly in all the possible calls.
- [ ] If there are more than one function returning this error code, all of them should be validated either as one test case or as multiple test cases.
       
## Acceptance Criteria
NA

## Implementation Details
Look at other test cases, and add a test case for the same.
Ex: check method `trying_to_register_statement_by_a_non_delegate_should_fail` which validates error code `UnauthorizedOperation`

## Mockups / Wireframes
NA

---- 

## Product Name
CORD

## Organization Name
Dhiway

## Domain
Blockchain

## Tech Skills Needed
Rust

## Mentor(s)
@amarts 

## Complexity
[Medium]

## Category
[Test]

## Sub Category
[Beginner friendly]

Sayanjones:
@vatsa287 I'm interested on this project. Can we discuss further?

rising0raj:
Hey, @vatsa287 I would like to work on this . I have gone through the code base can you assign this to me if it is open??

Sohajawdekar:
Hello @vatsa287 !
I am interested in contributing to this project, can you please assign me this ticket ?

summerpun2803:
@vatsa287 I would like to contribute to this project

VishnuAmit:
Hey @vatsa287 I would like to take on this project. Kindly assign it to me. 
Regards

PayalKumari10:
Hello @vatsa287 @amarts, I want to work on this project
Can you assign this issue to me?

atharva1910:
What is the point of creating all these issues and trying to engage the community if dhiway as an organization is itself not serious about the contributions. I have never seen such blatant disregard for the contributors time. Honesty just close all these issues and stop wasting peoples times.

VedantKhairnar:
Hello @PayalKumari10 
Any update on the contribution? Pls do let us know if you need any assistance.
Thanks.

Vaibhavsahu2810:
[...]

PR #542, opened 2024-12-21
Fixes: #318 Add test cases for PresentationDigestAlreadyAnchored error

Vaibhavsahu2810:
Fix for- #318 .

This PR adds a new unit test to the pallet-statement module to verify that attempting to add a presentation with the same digest to a statement that already has a presentation with that digest results in the expected error:
 Error::< Test >::PresentationDigestAlreadyAnchored.

The test covers the following scenarios:

Creates a statement.
Adds a presentation to the statement.
Attempts to add the same presentation (with the same digest) to the same statement again.
Asserts that the second add_presentation call fails with the expected error.
This test enhances the test coverage of the pallet-statement module and improves its overall quality.

### DIFF
--- a/pallets/statement/src/tests.rs
+++ b/pallets/statement/src/tests.rs
@@ -1296,3 +1296,66 @@ fn nonexistent_presentation_should_fail() {
 		);
 	});
 }
+
+#[test]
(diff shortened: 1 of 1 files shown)


vatsa287 (approved):
LGTM

PR #543, opened 2024-12-21
Fixes: #312 Add test cases for MaxEncodedSchemaLimitExceeded error

### LINKED ISSUE #312: [C4GT] Schema: Add tests for `MaxEncodedSchemaLimitExceeded`

Vaibhavsahu2810:
Fixes #312 
This PR adds a new unit test to the pallet-statement module to verify that the Schema::create function correctly handles schemas that exceed the maximum allowed encoded length.

The test covers the following scenarios:

Creates a schema with the maximum allowed encoded length.

Verifies that creating a schema with data exceeding the maximum length fails.
This test enhances the test coverage of the pallet-statement module and improves its robustness by ensuring that schema creation adheres to the defined size limits.


pallets/schema/src/tests.rs
```diff
@@ -314,3 +314,54 @@ fn test_schema_lookup() {
...
+		));
+		assert_ok!(Space::approve(RawOrigin::Root.into(), space_id, 10));
+
+		assert_ok!(Schema::create(
+			DoubleOrigin(author.clone(), creator.clone()).into(),
+			max_schema,
+			authorization_id.clone()
+		));
+
+		assert!(
```
vatsa287, line 353: Use `assert_err` with the error code instead of manual assertion.

vatsa287 (changes requested):

pallets/schema/src/tests.rs
Vaibhavsahu2810: Hello @vatsa287 ,when i am trying to use assert_err, panick get caused, the schema data can't even be converted to a BoundedVec because it exceeds the size limit, so we never get to the point of calling Schema::create
image

Vaibhavsahu2810:
@vatsa287 what's your thought on this?? is there any other way to do

PR #558, opened 2025-01-02
MaxKeyAgreementKeysExceeded

krishnamonga:
This test ensures that the MaxKeyAgreementKeysExceeded error is triggered when the number of key agreement keys exceeds the defined maximum limit during DID creation. It validates this behavior by attempting to exceed the limit and asserting the expected error response from both the key addition and DID creation processes. This confirms that the pallet correctly enforces the key agreement keys limit.

I added it

### DIFF
--- a/pallets/did/src/tests.rs
+++ b/pallets/did/src/tests.rs
@@ -2874,3 +2874,33 @@ fn check_invalid_signature_operation_verification() {
 		);
 	});
 }
+
+#[test]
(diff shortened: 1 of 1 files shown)


vatsa287:
@krishnamonga 
Refer Readme on contributing guidelines on the set of tests need to performed for the PR to pass. 
Also please update the description to include the issue number the PR fixes, Ex: `Fixes: #1000`.

PR #576, opened 2025-02-06
Collection, Registries and Entries

smohan-dw:


pallets/registry/Cargo.toml
```diff
@@ -0,0 +1,67 @@
+[package]
```
amarts, line 1: Is this registry code similar to pallet-registry which already there? Or any changes in logic? (other than removing tight coupling with namespace?)

pallets/registry/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+			registry::archive_registry::<T>(&registry_id, who.clone())?;
+			Self::deposit_event(Event::RegistryArchived { registry: registry_id, authority: who });
+			Ok(())
+		}
+
+		/// Restore registry
+		#[pallet::call_index(4)]
+		#[pallet::weight({10_000})]
+		pub fn restore(origin: OriginFor<T>, registry_id: RegistryIdentifierOf) -> DispatchResult {
+			let who = ensure_signed(origin)?;
```
amarts, line 199: Is it a TODO to look at delegate permission of the origin (`who`) here?

amarts (commented):
Looks fine.

One question is, if I want to use CORD (ie, in this case `pallet-registry`) directly without `doc_id`, and `doc_author_id` what would be the options?

amarts:
Merging this on a separate branch (with rc-0.9.5 as base), so few development efforts on this can continue, including these changes.

PR #584, opened 2025-03-03
identifier: Add verification runtime api for identifiers 

vatsa287:


primitives/identifier/src/lib.rs
```diff
@@ -118,3 +118,13 @@ impl<T: Config> IdentifierUpdate<IdentifierOf, IdentifierTypeOf, EventEntryOf, I
 		.map_err(|_| IdentifierError::MaxEventsHistoryExceeded) // Map DispatchError to your custom Error
 	}
 }
+
+impl<T: Config> Pallet<T> {
+	pub fn identifier_exists(id: &IdentifierOf) -> bool {
+		Identifiers::<T>::iter_prefix(id).next().is_some()
```
amarts, line 124: Does this return the data also? (ie, hash, block index and identifier?

primitives/identifier/src/lib.rs
```diff
@@ -118,3 +118,13 @@ impl<T: Config> IdentifierUpdate<IdentifierOf, IdentifierTypeOf, EventEntryOf, I
...
+	}
+
+	pub fn identifier_exists_by_type(id: &IdentifierOf, id_type: IdentifierTypeOf) -> bool {
```
amarts, line 127: Can the type be optional? (if the above call is not returning content).

primitives/identifier/src/lib.rs
amarts: Answering here myself. I see `bool` type as the return :-)

primitives/identifier/src/lib.rs
vatsa287: For unknown type `identifiers_exist` can be used. `identifier_exists_by_type` is faster when which pallet type is known.

amarts (approved):

PR #590, opened 2025-03-08
pallet-chain_space: Added test cases for Invalid Identifier

ritankarsaha:
This PR fixes issue #279 

## Implementation Approach

1.
`creating_space_identifier_with_invalid_input_should_fail`

`Setup`:
It sets up the testing environment using new_test_ext().execute_with(|| { … });, which initializes the mock runtime for our pallet.
`Invalid Input`:
The test creates an empty byte vector (vec![]). In a real scenario, the identifier function expects a valid byte sequence (for example, a hash of some space data). An empty input is considered invalid.
`Action`:
The test calls Ss58Identifier::create_identifier with the empty byte vector and specifies that it is to create a Space identifier by passing IdentifierType::Space.
`Assertion`:
The test asserts that this function call returns an error. If the function correctly detects the invalid input, it will not produce a valid identifier and instead return an error, which is exactly what the test checks for.

2.
`creating_authorization_identifier_with_invalid_input_should_fail`

`Setup`:
Similarly, it initializes the testing environment with new_test_ext().execute_with(|| { … });.
`Invalid Input`:
Again, the test uses an empty byte vector as the input, which is not acceptable for a valid identifier.
`Action`:
It calls Ss58Identifier::create_identifier, this time specifying IdentifierType::Authorization to simulate creating an authorization identifier.
`Assertion`:
The test checks that the function call returns an error. This ensures that if an invalid input is given when trying to create an authorization identifier, the function correctly fails.

## Local Testing

Test 1:- 
Screenshot 2025-03-08 at 3 08 21 PM

Test 2:- 

Screenshot 2025-03-08 at 3 09 06 PM

### DIFF
--- a/pallets/chain-space/src/tests.rs
+++ b/pallets/chain-space/src/tests.rs
@@ -1250,30 +1250,30 @@ fn add_delegator_should_fail_for_space_creator_delegating_themselves() {
 }
 
 #[test]
-    fn creating_space_identifier_with_invalid_input_should_fail() {
-        new_test_ext().execute_with(|| {
-            // An empty byte slice is assumed to be invalid.
-            let bad_input: Vec<u8> = vec![];
-            let result = Ss58Identifier::create_identifier(&bad_input, IdentifierType::Space);
-            assert!(
-                result.is_err(),
-                "Expected failure when creating a space identifier with invalid input"
-            );
-        });
-    }
+fn creating_space_identifier_with_invalid_input_should_fail() {
+	new_test_ext().execute_with(|| {
+		// An empty byte slice is assumed to be invalid.
+		let bad_input: Vec<u8> = vec![];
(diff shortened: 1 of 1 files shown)


ritankarsaha:
@vatsa287 are these changes in accordance? or do i need to make any more modifications to this PR?

ritankarsaha:
> @vatsa287 are these changes in accordance? or do i need to make any more modifications to this PR?

Hello, @vatsa287 Do i need to make any more modifications to this or is this good to go?

vatsa287:
@ritankarsaha It is good enough. Cannot merge right now, as we are making a release. This would be part of next release.

vatsa287 (approved):
LGTM

ritankarsaha:
> @ritankarsaha It is good enough. Cannot merge right now, as we are making a release. This would be part of next release.

Sure @vatsa287 😊

PR #597, opened 2025-03-17
move asset to asset-did (similar to other pallets)

amarts:


vatsa287 (approved):
LGTM.

If we are planning to bring this(or account based) into our runtime, we can update the name of the pallet reference from `Asset` to `AssetDid`. This would bring changes in tests, identifier, benchmarking, runtime.

PR #601, opened 2025-04-03
Add `pallet-profile` for new Identity management.

vatsa287:
### Pallet Profile

The current **Identity** pallet which is used for Identity management is tied up with an account and is therefore stateless. 
Watch cannot be established on the events associated to the Identity later. 
Also on the `rotation` of keys, the entire identity would be lost or have to be migrated to new key again to keep it stateful.

The new `pallet-identity` focuses on solving these issues entirely so it is much in line with our stack's needs. 
- Allows users or organisational accounts of CORD to create their identity on chain tied to a particular `CORD URI` based identifier. ( More on this later ) 
- Ownership transfer is made possible easily by just providing the new-public key which they want this Profile to be managed by here on.
- Activity or Event watching and broadcasting is made possible with the inclusion of the the Identifier based approach.
- Historical Keys can be provided for showing provenance from an external off-chain workers.

CORD URI:
This pallet uses new CORD URI based identifier which allows to extract the exact Network ID, Pallet the identifier belongs, helping to identify the route of the identifier in a multi-chain setup.

We can integrate this with our Registries, Entries, Collections and even for `myn.social` use cases to have persistent identities.

@amarts @smohan-dw

### DIFF
--- a/pallets/profile/src/lib.rs
+++ b/pallets/profile/src/lib.rs
@@ -1,0 +1,277 @@
+// This file is part of CORD – https://cord.network
+
+// Copyright (C) Dhiway Networks Pvt. Ltd.
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+// CORD is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// CORD is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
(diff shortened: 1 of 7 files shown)


pallets/profile/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+
+            ensure!(
+                !Profiles::<T>::contains_key(&profile_id),
+                Error::<T>::ProfileAlreadyExists
+            );
+
+            let profile = ProfileMetadataOf::<T> { latest_key: who.clone() }; 
+            Profiles::<T>::insert(&profile_id, profile);
+            AccountProfiles::<T>::insert(&who, &profile_id);
+
```
amarts, line 221: most of the times, lets insert the Storage entries after all validations happen. What if this `set_profile()` fails with `InvalidKeyPrefix` below. In that case, we will have many dangling profile identifier, which are not valid.

amarts (approved):

PR #618, opened 2025-04-17
entry: Add Hash to Identifier Lookup

vatsa287:
As the current identifiers are generated dynamically and considered as a token guaranteed to be on the chain, 
the apps/sdks would need to get the identifier generated from the content (digest) they sent which was ultimately used in part for creation of the identifier.

So create a new storage to have a lookup of identifier of registry entry based out of RegistryEntryHash and RegistryIdentifier.

In `Rust` we can iterate using `iter_prefix` for getting the identifier with registry-id being unknown.

But in `typescript` we need to fetch the entire storage and check for existence.
Example usage:
```
    const entries = await api.query.entry.hashToIdentifier.entries();
    const matches = entries
      .filter(([key]) => {
        const [key1] = key.args; 
        return key1.toHex() === tx_hash; 
      })
      .map(([key, value]) => {
        const [, registry_id] = key.args; 
        return {
          registry_id: registry_id.toHuman(), 
          registry_entry_id: value.isNone ? null : value.unwrap().toHuman(), 
        };
      });
 ```
 The above will be integrated with the demo script at SDK CORD.js

### DIFF
--- a/pallets/entry/src/lib.rs
+++ b/pallets/entry/src/lib.rs
@@ -127,7 +127,28 @@ pub mod pallet {
 	/// It maps Registry Entry Identifier to Registry Entry Details.
 	#[pallet::storage]
 	pub type RegistryEntries<T: Config> =
-		StorageMap<_, Blake2_128Concat, RegistryEntryIdOf, RegistryEntryDetailsOf<T>, OptionQuery>;
+		StorageMap<
+			_, 
+			Blake2_128Concat, 
+			RegistryEntryIdOf, 
+			RegistryEntryDetailsOf<T>, 
+			OptionQuery
+		>;
+
+	/// Storage to map for Entry hashes to corresponding Registry Identifiers.
+	/// It being a storage double-map will have the Registry Entry Hash and the Registry ID
+	/// as the key, whereas the value resulted is the Registry Entry Identifier.
+    #[pallet::storage]
+    pub type HashToIdentifier<T> = 
+        StorageDoubleMap<
+            _,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: pallets/entry/src/lib.rs [hunk 1]
```diff
@@ -127,7 +127,28 @@ pub mod pallet {
...
+			RegistryEntryDetailsOf<T>, 
+			OptionQuery
+		>;
+
+	/// Storage to map for Entry hashes to corresponding Registry Identifiers.
+	/// It being a storage double-map will have the Registry Entry Hash and the Registry ID
+	/// as the key, whereas the value resulted is the Registry Entry Identifier.
+    #[pallet::storage]
+    pub type HashToIdentifier<T> = 
+        StorageDoubleMap<
```
amarts, line 143: Why cant we do single Hash to Entry? Are we allowing same hash to be anchored by different registries?

Also, if I keep registryIdentifier empty, I get all the options right? IMO, that is good enough.

amarts:
Don't we need insert to this table?

pallets/entry/src/lib.rs
vatsa287: Yes same hash can be inserted by different users to different registries. So have added registry-id. 

It is entirely possible to get the identifier for the entry hash `without` knowing the registry-id, as shown in the script. I will demo this in the SDK for reference later.

vatsa287:
> Don't we need insert to this table?

Yes the insertion is already done at `create` and `update`

amarts (approved):

PR #620, opened 2025-04-18
registry: Split registry creation for native and cyra

vatsa287:
Splitting the current union of creation of a registry into 

- create: For creating registry natively on-chain.
- create-store: For anchoring creation of registry made on `cyra` onto `cord`

^ @amarts @smohan-dw


amarts (approved):
In line with our discussion. This should be captured in `specs/` repository

PR #635, opened 2025-05-14
registry: Fix benchmarks & update tests for profile 

vatsa287:
Ensure compiling code for benchmark production with latest profile embedded codebase.


amarts (commented):
LGTM

PR #640, opened 2025-05-29
entry, weave: Add runtime api for verifying digest 

vatsa287:
- Verify if the digests exits, if yes return the registry-entry-id. It shall be in `hex`. Verifiers are needed to convert to `base58`.
- If the `registry-id` is not given, it iterates through all registry-id, else just a O(1) operation.

### DIFF
--- a/pallets/entry/src/lib.rs
+++ b/pallets/entry/src/lib.rs
@@ -66,6 +66,7 @@ pub use types::RegistryEntryDetails;
 pub use cord_primitives::StatusOf;
 use pallet_profile::ProfileIdOf;
 use pallet_registry::{Permissions, RegistryIdentifierOf};
+// use alloc::string::String;
 
 #[frame_support::pallet]
 pub mod pallet {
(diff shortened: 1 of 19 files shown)


runtimes/weave/src/lib.rs
```diff
@@ -2534,6 +2537,12 @@ impl_runtime_apis! {
 			vec![]
 		}
 	}
+
+	impl crate::registry_entry_api::RegistryEntryApi<Block, Hash, RegistryIdOf> for Runtime {
```
smohan-dw: Can't we use the Ss58Identifier type here instead of RegistryIdOf. Good to keep runtime out of pallet-derived types.

pallets/entry/src/lib.rs
```diff
@@ -601,6 +602,16 @@ pub mod pallet {
 	}
 }
 
+// pub trait Ss58Encoding {
```
smohan-dw: remove this.

pallets/entry/src/lib.rs
```diff
@@ -615,4 +626,21 @@ impl<T: Config> Pallet<T> {
 			.map_err(|_| Error::<T>::StateUpdateFailed)?;
 		Ok(())
 	}
+
+	pub fn verify_digest(
```
smohan-dw, line 621: Optimise this. for eg:
```
pub fn verify_digest(
    digest: T::Hash,
    registry_id: Option<RegistryIdOf>,
) -> Result<Option<RegistryEntryIdOf>, Error<T>> {
    let result = if let Some(reg_id) = registry_id {
        // Direct lookup: O(1)
        HashToIdentifier::<T>::get(&digest, reg_id)
    } else {
        // Prefix lookup: stop at first match (O(1) in practice, instead of O(n))
        HashToIdentifier::<T>::iter_prefix(&digest)
            .next()
            .map(|(_reg_id, entry_id)| entry_id)
    };

    Ok(result)
}
```

vatsa287: Updated

amarts (approved):

PR #646, opened 2025-07-10
Update to 9800, so weave-testnet can be upgraded

vatsa287:


amarts (approved):

runtimes/weave/src/lib.rs
```diff
@@ -157,7 +157,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: alloc::borrow::Cow::Borrowed("weave"),
 	impl_name: alloc::borrow::Cow::Borrowed("dhiway-cord"),
 	authoring_version: 0,
-	spec_version: 9700,
+	spec_version: 9800,
```
smohan-dw, line 160: The spec versions should also align with the cargo versions. If we are updating this, please update the cargo version as well

